Issue #16, opened 2017-04-28
Sync word issue


tigoe:
Sync word checking appears to be inconsistent. (I'm using a variation of the LoRaReceive sketch here, not the callback sketch). What follows is a set of collected findings and questions, but unfortunately not a solution (sorry).

Using two radios, I set the sync word of one to 0xAA and the other variable. I ran through all values from 0x01 to 0xFF. The following "matched" with 0xAA:

`a8, a7, a6, a5, a4, a3, a2, a1, a0, 9d, a8, a9, aa, ab, ac, ad, ae, af, b0, b3, a2, a3, a4, a5, a6, a7, a8, a9, aa, ab, ac, ad, ae, af, b0, b3, b4, b5, b9, bc, c3, c5, c6, c9, ca, d3, d6, d7, d9, e3, e8, ea, f0, f7, 13, 1a, 2c, 2f, 33, 36, 39, 3a, 43, 44, 45, 46, 4f, 5d, 67, 69, 6f, 70, 7a, 7c, 7f, 88, 89, 8e, 94, 95, 9e, a3, a4, a5, a6, a7, a8, a9, aa, ab, ac, ad, ae, af, b0`

It seems like the device should be in sleep mode when you set the sync word, so perhaps it would make sense to wrap the writeRegister() in setSyncWord() in a call to sleep mode, then return the radio to its normal mode before the function is over.  Also, "The bit synchronizer must also be activated in Continuous mode (automatically done in Packet mode)" (from the [Semtech datasheet](http://www.semtech.com/images/datasheet/sx1276.pdf)), which makes me wonder if sync word detection doesn't work in single receive mode? The HopeRF datasheet is unclear on this, and doesn't even detail the LoRa mode registers.

It is worth noting somewhere in the docs that single receive mode is intended for battery operation, and continuous receive mode consumes more power, and is intended for non-battery operation.

Will test more when I've got the hardware in hand again.

sandeepmistry:
> It seems like the device should be in sleep mode when you set the sync word, so perhaps it would make sense to wrap the writeRegister() in setSyncWord() in a call to sleep mode, then return the radio to its normal mode before the function is over.

Oops, I missed that during my testing, so it was probably not using the desired sync work value. I'll open a PR for the change.


> Also, "The bit synchronizer must also be activated in Continuous mode (automatically done in Packet mode)" (from the Semtech datasheet), which makes me wonder if sync word detection doesn't work in single receive mode? 

[...]

PR #17, opened 2017-04-28
Added SyncWordTest

tigoe:
Here's what I've got so far as a test on syncWord


examples/LoRaDuplex/LoRaDuplex.ino
```diff
@@ -0,0 +1,101 @@
...
+  Serial.begin(9600);                   // initialize serial
+  Serial.println("LoRa Duplex");
+  LoRa.setPins(csPin, resetPin, irqPin);// set CS, reset, IRQ pin
+
+  if (!LoRa.begin(915E6)) {             // initialize ratio at 915Mhz
+    Serial.println("LoRa init failed. Check your connections.");
+    while (true);                       // if failed, do nothing
+  }
+
+   LoRa.setTimeout(10);                  // set Stream timeout of 10ms
```
sandeepmistry: Any reason not to set this to 0? Since the received packet is buffered, I think this should be fine.

We can even make the LoRa library set a lower default timeout internally ... thoughts?

examples/LoRaDuplexCallback/LoRaDuplexCallback.ino
```diff
@@ -0,0 +1,100 @@
...
+  LoRa.write(msgCount);                 // add message ID
+  LoRa.write(outgoing.length());        // add payload length
+  LoRa.print(outgoing);                 // add payload
+  LoRa.endPacket();                     // finish packet and send it
+  msgCount++;                           // increment message ID
+}
+
+
+void onReceive(int packetSize) {
+  if (packetSize == 0) return;          // if there's no packet, return
```
sandeepmistry: In the callback case, I don't think the callback will be called with 0 packet size, is this here because of the non-callback example above?

test/LoRaDumpRegisters/LoRaDumpRegisters.ino
```diff
@@ -0,0 +1,26 @@
+/*
+  LoRa register dump
```
sandeepmistry: Is this a duplicate of https://github.com/sandeepmistry/arduino-LoRa/pull/17/files#diff-3d27af960cba75a3af0e3fc42ca29bb2?

examples/LoRaDumpRegisters/LoRaDumpRegisters.ino
```diff
@@ -0,0 +1,26 @@
...
+void setup() {
+  while (!Serial);
+  Serial.begin(9600);               // initialize serial
+  Serial.println("LoRa Receiver");
+  LoRa.setPins(7, 6, 1);            // set CS, reset, IRQ pin
+  if (!LoRa.begin(915E6)) {         // initialize ratio at 915Mhz
+    Serial.println("LoRa init failed. Check your connections.");
+    while (true);                   // if failed, do nothing
+  }
+  LoRa.sleep();
```
sandeepmistry: I think this is missing from the docs, `LoRa.sleep()` needs to be called before `LoRa.dumpRegisters(...)` right?

test/LoRaDuplexAutoContinuous/LoRaDuplexAutoContinuous.ino
```diff
@@ -0,0 +1,129 @@
...
+  if (millis() - lastSend > 500) {
+    message = "sender's syncWord:\t";  // send a message; tab automatically
+    message += String(syncWord, HEX);  // separates the sync word into its own column
+    sendMessage(message);
+    lastSend = millis();
+  }
+  //  Change the sync word once every 10 messages:
+  //if(messageId % 10 == 0 && change == 1) {
+  //  LoRa.sleep();
+  //  LoRa.setSyncWord(syncWord++);
```
sandeepmistry: Another thing that needs to be in the docs right, `LoRa.sleep()` needs to be called before ` LoRa.setSyncWord(...)` ...

examples/LoRaDuplex/LoRaDuplex.ino
```diff
@@ -0,0 +1,101 @@
+/*
+  LoRa Duplex communication
```
tigoe: The test example was just to show the work on the sync word issue. I think it's too complex for an example, you can delete that whole folder when you're done with it.

examples/LoRaDuplex/LoRaDuplex.ino
tigoe: My experience with the Stream timeouts has been that timeout(0) tends to miss bytes on long strings, so I default to 10. But since the packet is buffered, it might be alright.

examples/LoRaDuplexCallback/LoRaDuplexCallback.ino
tigoe: It's because of the non-callback case, and I just copy-pasted, but you're right, it's not necessary.

test/LoRaDumpRegisters/LoRaDumpRegisters.ino
tigoe: Correct. Sorry about that.

examples/LoRaDumpRegisters/LoRaDumpRegisters.ino
tigoe: I think you can dump without sleep. You can't write without sleep to many of the control regs, though.

test/LoRaDuplexAutoContinuous/LoRaDuplexAutoContinuous.ino
tigoe: Definitely.  In fact, I think it'd be best just to wrap that internally. I think you already PRed that though.

sandeepmistry:
Closing in favour of #24.

examples/LoRaDuplex/LoRaDuplex.ino
sandeepmistry: I've made this change in PR #24.

examples/LoRaDumpRegisters/LoRaDumpRegisters.ino
sandeepmistry: I've removed the sleep in PR #24.

test/LoRaDuplexAutoContinuous/LoRaDuplexAutoContinuous.ino
sandeepmistry: Yes, doesn't seem to help though, see new notes in #16.

PR #18, opened 2017-05-07
Backup and restore opcode while setting sync word

sandeepmistry:
As discussed in https://github.com/sandeepmistry/arduino-LoRa/pull/17#discussion_r114056252.

@tigoe I'm still seeing some weirdness with this change, sometime it doesn't look like the new sync word is used on the TX side.


sandeepmistry:
Closing this for now based on: https://github.com/sandeepmistry/arduino-LoRa/issues/16#issuecomment-304542088

PR #59, opened 2017-11-11
Fix SNR register

YannikW:
I noticed strange behaviour with the SNR value (sometimes jumping from 10 to 20dB). Also I notices 100% packet loss with +10dB SNR (what can't be true with my understanding of RSSI and SNR).

Datasheet shows SNR register is 0x19, not 0x1b. 0x1b is raw rssi.

After changing the register I got *realistic* values for SNR.

image


sandeepmistry:
@YannikW great find! Thanks for submitting this PR :)

Issue #56, opened 2017-11-04
saving pins ss and reset


universam1:
Pins on ESP8266 platform are expensive. 
Reading through your great library I assume that the pins ss and reset could be omitted- optionally! 

SS could be pulled down permanently if single SPI device. 
Reset could be wired to ESP8266 reset pin and get resetted at boot. 
A little support by library would be required I could provide and test. 
What do you think about?

sandeepmistry:
@universam1 please submit a pull request to continue the discussion.

universam1:
Roger, I will test first, ordered hardware.

universam1:
Okay I tested this whole setup, and must step back from suggesting to save pin `ss`, it does not work in my setup ESP8266 with RA-02 reliable, so removed it from MR.

Nevertheless, omitting the `reset` pin by hardwiring to the MCU works just perfect, for me there is no use to spend an IO for this. 
Feel free to merge @sandeepmistry

sandeepmistry:
Closed via #60.

PR #60, opened 2017-11-13
saving reset pin

universam1:
should cover #56

Reset could be wired to ESP8266 reset pin and get resetted at boot.

### DIFF
--- a/API.md
+++ b/API.md
@@ -32,12 +32,11 @@ LoRa.setPins(ss, reset, dio0);
 
 This call is optional and only needs to be used if you need to change the default pins used.
 
-### Saving pin
+#### No MCU controlled reset pin
 
-To save further pins one could connect the Reset pin of the MCU with reset pin of the radio thus resetting only during startup.
+To save further pins one could connect the reset pin of the MCU with reset pin of the radio thus resetting only during startup.
 
 * `reset` - set to `-1` to omit this pin
-
 
 ### Set SPI Frequency
 
--- a/API.md
+++ b/API.md
@@ -32,6 +32,14 @@ LoRa.setPins(ss, reset, dio0);
 
 This call is optional and only needs to be used if you need to change the default pins used.
 
+### Saving pins (Experimental)
+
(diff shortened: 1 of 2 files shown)


universam1:
This PR @sandeepmistry has been tested with ESP8266 and AI-Thinker RA-01 and works perfectly!

src/LoRa.cpp
```diff
@@ -66,17 +66,20 @@ int LoRaClass::begin(long frequency)
...
-  digitalWrite(_reset, LOW);
-  delay(10);
-  digitalWrite(_reset, HIGH);
-  delay(10);
-
   // set SS high
   digitalWrite(_ss, HIGH);
 
+  if (_reset != -1)
+  {
```
sandeepmistry: Could you please make this bracket consistent with the rest of the library.

src/LoRa.cpp
```diff
@@ -502,7 +505,6 @@ void LoRaClass::writeRegister(uint8_t address, uint8_t value)
 uint8_t LoRaClass::singleTransfer(uint8_t address, uint8_t value)
 {
   uint8_t response;
-
```
sandeepmistry: Could you please undo this change?

API.md
```diff
@@ -32,6 +32,13 @@ LoRa.setPins(ss, reset, dio0);
 
 This call is optional and only needs to be used if you need to change the default pins used.
 
+### Saving pin (tested on ESP8266)
```
sandeepmistry: Even though it's only been tested on ESP8266, I think we can leave out the text in the brackets.

universam1: fine with me

src/LoRa.cpp
universam1: np - my stylecheck is really enemy with your file... I dont mind.

src/LoRa.cpp
universam1: sure

universam1:
requests applied

sandeepmistry:
@universam1 thanks!

PR #61, opened 2017-11-16
stabilize ESP8266 watchdog

### LINKED ISSUE #38: ESP32/ESP8266 Issues
Hi,  
I have been working with your LoRaDuplex sketch and I ran into some issues.
I was able to get it running with a Leonardo with a Dragino LoRa radio, I was also able to get it running with a M0 basic Feather and LoRa Feather. When I tried to get it working with an ESP32 I could never get the radio initialized, and when I tried to get it running on an ESP8266 it would just keep resetting. 
      Soft WDT reset
      ctx: cont
      sp: 3ffef9f0 end: 3ffefbd0 offset: 01b0
      >>>stack>>>
      3ffefba0: 3fffdad0 3ffeea78 3ffeeb7c 40201f65
      3ffefbb0: feefeffe 00000000 3ffeeb98 4020331c
      3ffefbc0: feefeffe feefeffe 3ffeebb0 40100114
      <<<stack<<<
I know the hardware works, but would there be an issue with the ESP32/ESP8266.  I wouldn't think so but this has me stumped.
Any thoughts?

Thanks,

universam1:
solve the watchdog timeout issue on ESP8266
fixes #38

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -144,8 +144,9 @@ int LoRaClass::endPacket()
   writeRegister(REG_OP_MODE, MODE_LONG_RANGE_MODE | MODE_TX);
 
   // wait for TX done
-  while ((readRegister(REG_IRQ_FLAGS) & IRQ_TX_DONE_MASK) == 0)
+  while ((readRegister(REG_IRQ_FLAGS) & IRQ_TX_DONE_MASK) == 0) {
     yield();
+  }
 
   // clear IRQ's
   writeRegister(REG_IRQ_FLAGS, IRQ_TX_DONE_MASK);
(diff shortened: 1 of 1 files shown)


src/LoRa.cpp
```diff
@@ -144,7 +144,11 @@ int LoRaClass::endPacket()
   writeRegister(REG_OP_MODE, MODE_LONG_RANGE_MODE | MODE_TX);
 
   // wait for TX done
-  while((readRegister(REG_IRQ_FLAGS) & IRQ_TX_DONE_MASK) == 0);
+  while((readRegister(REG_IRQ_FLAGS) & IRQ_TX_DONE_MASK) == 0)
+#ifdef ESP8266
```
sandeepmistry: I think we should remove the `#ifdef` and call yield on all architectures.

universam1: I wasn't sure if this is working on all platform, thats why I set the guard. If you think so, I prefere!

universam1:
request applied!

sandeepmistry:
@universam1 thanks!

PR #62, opened 2017-11-17
non blocking functions added

universam1:
the `endPacket()` is a blocking function that can take over 6 seconds depending on the radio settings. Here is a non blocking version added, tested successfully on ESP8266. Note the grace time required for the radio to properly update the registers.

### DIFF
--- a/API.md
+++ b/API.md
@@ -71,7 +71,7 @@ LoRa.beginPacket(implicitHeader);
 
  * `implicitHeader` - (optional) `true` enables implicit header mode, `false` enables explicit header mode (default)
 
-Returns `1` if radio is ready to transmit, `0` if busy or on failure. Consider your code responsible for testing first before actually writing to radio.
+Returns `1` if radio is ready to transmit, `0` if busy or on failure.
 
 ### Writing
 
@@ -95,13 +95,15 @@ Returns the number of bytes written.
 
 ### End packet
 
-End the sequence of sending a packet. Set `async` to `true` in order to switch to non-blocking mode.
+End the sequence of sending a packet.
 
 ```arduino
-LoRa.endPacket(bool async)
+LoRa.endPacket();
+
+LoRa.endPacket(async);
 ```
- * `async` - (optional) `true` enables non-blocking mode, `false` returns after transmission (default)
-  
+ * `async` - (optional) `true` enables non-blocking mode, `false` waits for transmission to be completed (default)
+
 Returns `1` on success, `0` on failure.
 
 ## Receiving data
--- a/examples/LoRaSenderNonBlocking/LoRaSenderNonBlocking.ino
+++ b/examples/LoRaSenderNonBlocking/LoRaSenderNonBlocking.ino
@@ -1,0 +1,40 @@
+#include <SPI.h>
+#include <LoRa.h>
+
+int counter = 0;
+
+void setup()
+{
+  Serial.begin(9600);
+  while (!Serial)
+    ;
+
+  Serial.println("LoRa Sender");
+
+  if (!LoRa.begin(915E6))
+  {
+    Serial.println("Starting LoRa failed!");
+    while (1)
+      ;
+  }
+}
+
+void loop()
+{
+  if (LoRa.isTransmitting())
+  {
+    delay(100);
+    Serial.print('w');
+    return;
+  }
+  Serial.print("\nSending packet: ");
+  Serial.println(counter);
+
+  // send packet
+  LoRa.beginPacket();
+  LoRa.print("hello ");
+  LoRa.print(counter);
+  LoRa.endPacketasync();
+
+  counter++;
+}
--- a/examples/LoRaSenderNonBlocking/LoRaSenderNonBlocking.ino
+++ b/examples/LoRaSenderNonBlocking/LoRaSenderNonBlocking.ino
@@ -3,40 +3,33 @@
 
 int counter = 0;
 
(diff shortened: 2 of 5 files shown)


sandeepmistry:
@universam1 thanks for submitting!

Have you seen these API's in any other packet based Arduino libraries? Just wondering if there's existing API's we can leverage.

universam1:
Well, I would say it is actually more of a standard then exceptional to support non blocking calls (http://bfy.tw/FAaI)
There are even mcu's like ESP8266 that are built in an **assumption** that calls are
non blocking to happen. That's why my other PR #61 I would rather call a dirty workaround then a proper implementation.

So out of my head lets mention 
https://github.com/sadr0b0t/stepper_h
https://github.com/arduino/Arduino/pull/1240
http://www.epyon.be/2016/05/22/non-blocking-ethernet-example-teensy https://github.com/arduino/Arduino/issues/1476
https://github.com/jrullan/neotimer
https://github.com/olewolf/DHT_nonblocking
 and numerous other ones I came across.

Don't get me wrong, I consider your library is great, but non blocking functions are mandatory for me and turn a great library to professional. 
Of course if you like different function names, no problem!

universam1:
Would like to get this moving and reach an agreement if you @sandeepmistry  and @morganrallen  are happy with this implementation or rather overload the existing function with an async parameter?

sandeepmistry:
I'd like to take a look at how other existing Stream based Arduino libraries handle async. The links from https://github.com/sandeepmistry/arduino-LoRa/pull/62#issuecomment-346287239 are not necessarily Stream based.

Maybe nothing exists yet?

universam1:
I agree a streaming support might be desirable at the long run! Though this would be a bigger step. 
Following [DevOps style](https://en.wikipedia.org/wiki/Incremental_build_model) it is normally preferable to release often but smaller.
If a feature can be implemented without breaking backward compatibility this can be considered good to deploy.
But of course it depends on your philosophy for this project!

sandeepmistry:
Some API ideas:

* have either `endPacket` (or `beginPacket`) have an optional API for async mode
* then `beginPacket` fails (returns 0), if a packet is currently been sent by the radio.

universam1:
@sandeepmistry Implemented:
* `endPackt` has the optional parameter `bool async`
* and `beginPacket` fails if currently sending

Please review if that is fine, thanks

MauroMombelli:
@sandeepmistry the Wire library iunder the hood uses the TWI library, directly provided by atmel. Te TWI exposes all non blocking operation and is quite simple to use (you can use the Wire as example after all)

i proposed a fix to the arduino libr back in 2014.... https://github.com/arduino/Arduino/pull/2489

examples/LoRaSenderNonBlocking/LoRaSenderNonBlocking.ino
```diff
@@ -0,0 +1,42 @@
...
+
+int counter = 0;
+
+void setup()
+{
+  Serial.begin(9600);
+  while (!Serial)
+    ;
+
+  Serial.println("LoRa Sender");
```
sandeepmistry: Maybe adding "non-blocking" to the title here?

src/LoRa.h
```diff
@@ -22,7 +22,8 @@ class LoRaClass : public Stream {
   void end();
 
   int beginPacket(int implicitHeader = false);
-  int endPacket();
+  int endPacket(bool async = false);
+  bool isTransmitting();
```
sandeepmistry: I'd prefer if this API was private for now.

MauroMombelli:
@sandeepmistry please use 0 as SUCCESS, and other value as failure, unix (and rest of the codebase) stile

src/LoRa.h
morganrallen: How is this meant to be private if the current `endPacket` is public?

morganrallen: Does this mean, leave `int endPacket()` in public and move `int endPacket(bool async = false)` private? Don't know C++ classes well enough to understand what this will accomplish.

morganrallen:
I've checked this PR out and it appears to work so I'm inclined to merge it as it's an important feature. I'll address the comments and merge this evening if there are no objections.

@sandeepmistry @universam1

universam1:
fine with me 👍

src/LoRa.h
sandeepmistry: @morganrallen I was thinking `isTransmitting()` should be private for now.

With the current PR`beginPacket(...)`, fails if there is a transmission in progress.

sandeepmistry: @morganrallen any thoughts on the above?

morganrallen: Ah right, thanks makes sense, missed this comment last week.

universam1:
@morganrallen @sandeepmistry updated the PR accordingly, LGTM

API.md
```diff
@@ -103,6 +103,34 @@ LoRa.endPacket()
 
 Returns `1` on success, `0` on failure.
 
+## Sending data - non blocking
```
sandeepmistry: I don't think we should have a totally new section for non-blocking, rather modify the existing sending section, since there is really just a new parameter to `endPacket` and `beginPacket` has another reason to fail.

@morganrallen any thoughts on this?

universam1:
@sandeepmistry @morganrallen I've changed the API accordingly and removed the additional part merging now.
Please though LGTM, our review processes take way to long!

Issue #77, opened 2017-12-17
Support +20 dBm for SX1276


dontsovcmc:
It's will be good to add +20 dBm mode for SX1276 board and written in manual http://www.semtech.com/images/datasheet/sx1276.pdf
```
see 5.4.3. High Power +20 dBm Operation
```
I have 2 boards: https://www.ebay.com/itm/LoRa32u4-II-Lora-LiPo-Atmega32u4-SX1276-HPD13-868MHZ-EU-Antenna/282623601184

```
settings:
    LoRa.setSignalBandwidth(31.25E3); 
    LoRa.setSpreadingFactor(12); 
    LoRa.setCodingRate4(5);
    LoRa.enableCrc();
```

I check board current when transmitting at 17dBm and 20 dBm settings (as read in manual):

17 dBm:
` LoRa.setTxPower(17);`
current receive mode: 23mA
current transmit 1 byte with crc: 66mA (~4sec), 73mA (~2 sec)

20 dBm:
```
    writeRegister(REG_PA_CONFIG, PA_BOOST | (17 - 2));
    writeRegister(REG_PA_DAC, 0x87);  //turn on 3rd amplifier, see 5.4.3
    writeRegister(REG_OCP, 0x20 | 18); //increace power protection to 150mA RegOcp, see 5.4.4
```
current receive mode: 23mA
current transmit 1 byte with crc: 92mA (~4sec), 98mA (~2 sec)

I also see ~2 RSSI difference on receiver between 17 and 20 modes. 
I hadn't measure distance, sorry.

p.s. to turn off 20 dBm call:
```
    writeRegister(REG_PA_DAC, 0x84); //turn off 3rd amplifier, see 5.4.3
    writeRegister(REG_OCP, 0x20 | 0x0B); //RegOcp default
```

sandeepmistry:
Hi @dontsovcmc, please submit a pull request for the suggested changes.

Sukuri:
@dontsovcmc is this can be used in RFM96? and can I know what is the RFM96 PA_BOOST pin?, I'm confused about this pin because I didn't get it in datasheet

dontsovcmc:
I don't read your manual. 
 RFM96 should be the same  & pin should be soldered on board.

halukmy:
@dontsovcmc  add 1278 support  too

Sivick:
Hola buenas, estoy experimentando con este libreria y estoy midiendo el alcance maximo

[...]

PR #79, opened 2017-12-20
sx127x 20dBm support

dontsovcmc:
https://github.com/sandeepmistry/arduino-LoRa/issues/77

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -10,6 +10,7 @@
 #define REG_FRF_MID              0x07
 #define REG_FRF_LSB              0x08
 #define REG_PA_CONFIG            0x09
+#define REG_OCP                  0x0b
 #define REG_LNA                  0x0c
 #define REG_FIFO_ADDR_PTR        0x0d
 #define REG_FIFO_TX_BASE_ADDR    0x0e
@@ -31,6 +32,7 @@
 #define REG_SYNC_WORD            0x39
 #define REG_DIO_MAPPING_1        0x40
 #define REG_VERSION              0x42
+#define REG_PA_DAC               0x4d
 
 // modes
 #define MODE_LONG_RANGE_MODE     0x80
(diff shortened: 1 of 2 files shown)


bkeevil:
Looks good to me

wero1414:
I do not see anything wrong with this improvement, but i would suggest to not duplicate functions, cause this library contains already one function to change the power during Tx so, i would suggest to add the function  set20dBm_sx127x() as an option inside the setTxPower() function before its if statement about the PA_OUTPUT_RFO_PIN (https://github.com/sandeepmistry/arduino-LoRa/blob/5d6a7a398a03cdf0f97b0ce8a2c7c8ace8597868/src/LoRa.cpp#L342)

dontsovcmc:
Good idea!
Also, I think it's better call setOCP_sx127x() (current limit) before writeRegister(REG_PA_DAC, ..).

dontsovcmc:
The question: 
should we change OCP inside setTxPower function, or we should add notice for user "you must call setOCP(140); before setTxPower(20) ? 

```
void LoRaClass::setTxPower(int level, int outputPin)
{
  if (PA_OUTPUT_RFO_PIN == outputPin) {
    // RFO
    if (level < 0) {
      level = 0;
    } else if (level > 14) {
      level = 14;
    }

    writeRegister(REG_PA_CONFIG, 0x70 | level);
  } else {
    
    if (level > 17) {
      writeRegister(REG_PA_DAC, 0x87); //Value for High Power
      setOCP(140);
    } else {
      writeRegister(REG_PA_DAC, 0x84); //Default value PA_HF/LF or +17dBm
      setOCP(100);
    }
    
    // PA BOOST
    if (level < 2) {
      level = 2;
    } else if (level > 17) {
      level = 17;
    }

    writeRegister(REG_PA_CONFIG, PA_BOOST | (level - 2));
  }
}
```

halukmy:
currently not supports 20dbm?

wero1414:
Currently this library allows you only to transmit with a maximun of 17 dBm

wero1414:
@dontsovcmc i think that the better way is to change inside the OCP directly

dontsovcmc:
@sandeepmistry, what are you think about this improvments?
1. should we call setOCP() inside setTxPower()?

remark: OCP - current protection limit.

sandeepmistry:
@dontsovcmc here is my thoughts on what I'd like to see:

* no changes to the current public facing API's
* use the current public facing `LoRa.setTxPower(int level, int outputPin)`
  * if the `outputPin` is set to `PA_OUTPUT_RFO_PIN`: leave the current behaviour
  * if the `outputPin` is set to `PA_OUTPUT_PA_BOOST_PIN `:
    * if `level` is between 2 to 17: use the current behaviour, **and also set OCP and PA DAC accordingly**
    * if `level` is between 18 to 20: support these new ones by setting OCP, PA DAC and PA CONFIG accordingly.

@dontsovcmc @wero1414 any thoughts on the above?

wero1414:
Agree if you filter in agreement the level of transmission power you can set the correct configuration to control de current, for me its a real good improvement.

PR #97, opened 2018-02-01
Fix integer warnings

Rotzbua:
e.g.:
LoRaDuplex.ino:47:33: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

### DIFF
--- a/examples/LoRaDuplex/LoRaDuplex.ino
+++ b/examples/LoRaDuplex/LoRaDuplex.ino
@@ -23,8 +23,8 @@ String outgoing;              // outgoing message
 byte msgCount = 0;            // count of outgoing messages
 byte localAddress = 0xBB;     // address of this device
 byte destination = 0xFF;      // destination to send to
-long lastSendTime = 0;        // last send time
-int interval = 2000;          // interval between sends
+unsigned long lastSendTime = 0;        // last send time
+unsigned int interval = 2000;          // interval between sends
 
 void setup() {
   Serial.begin(9600);                   // initialize serial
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@Rotzbua thanks for submitting, what board are you building for?

When I compile for Arduino Uno on IDE 1.8.5 I don't see any warnings.

Rotzbua:
I usually switch the arduino ide to the highest warning level, then you will get the warning.

The `(byte)LoRa.read` is just for sanity, because `int` type is returned and do not fit into byte.

sandeepmistry:
> I usually switch the arduino ide to the highest warning level, then you will get the warning.

Thanks I missed that, for the Uno I now get (without this PR):

```
/Users/smistry/Documents/Arduino/libraries/LoRa/examples/LoRaDuplex/LoRaDuplex.ino: In function 'void loop()':
/Users/smistry/Documents/Arduino/libraries/LoRa/examples/LoRaDuplex/LoRaDuplex.ino:47:31: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (millis() - lastSendTime > interval) {
                               ^
```

> The (byte)LoRa.read is just for sanity, because int type is returned and do not fit into byte.

@tigoe any feedback on those changes? Do you think casting to `byte` is unfriendly to beginners?

Floessie:
> Do you think casting to `byte` is unfriendly to beginners?

Those explicit casts to `byte` on assignment aren't necessary. So the question shouldn't arise. Just my 2¢.

Best,
Flössie

Rotzbua:
Normaly I use clion, it always warns about implicit type conversation. Implicit type conversations are often reasons for bugs, so I personally do not like them.

It seems that `-Wall` do not include `-Wconversion` to enable the warning.

Floessie:
> Implicit type conversations are often reasons for bugs, so I personally do not like them.

Fair enough.

> It seems that `-Wall` do not include `-Wconversion` to enable the warning.

That's for a [reason](https://gcc.gnu.org/wiki/NewWconversion). Quoting:

> Implicit conversions are very common in C. This tied with the fact that there is no data-flow in front-ends (see next question) results in hard to avoid warnings for perfectly working and valid code. Wconversion is designed for a niche of uses (security audits, porting 32 bit code to 64 bit, etc.) where the programmer is willing to accept and workaround invalid warnings. Therefore, it shouldn't be enabled if it is not explicitly requested.

Issue #83, opened 2018-01-09
LoRa.onReceive fails on Heloc_WiFi_LoRA_32 (ESP32)


qbzzt:
I have two of these boards, one a sender and one a receiver.

When I use polling with the receiver ([code is here](https://github.com/qbzzt/IoT/blob/master/201801/Hay_Bale_Fire/05_LoRa_Receive.ino)), I lose a few packets from time to time. [But when I try to use onReceive, the function is never called](https://github.com/qbzzt/IoT/blob/master/201801/Hay_Bale_Fire/LoRa_onReceive_Fail).

Any idea why? I set the IRQ to the correct pin, I'm pretty sure ([26, as shown here](https://cdn.hackaday.io/files/269911154782944/Heltec_WIFI-LoRa-32_DiagramPinout.jpg)).

Thanks, Ori

qbzzt:
Additional information. When I try to bypass the LoRa library the ISR still doesn't get called

>   attachInterrupt(digitalPinToInterrupt(LORA_SPI_PINS.irq), 
>     onReceive, 
>     CHANGE);

sandeepmistry:
Can you try this sketch:

```arduino
void setup() {
  Serial.begin(115200);

  attachInterrupt(digitalPinToInterrupt(26), onRise, RISING);
}

void loop() {
  delay(100);
}

void onRise() {
  Serial.println("onRise");
}
```

Then manually attach a wire on pin 26 and switch between GND and 3V3. If nothing prints out it's mostly likely an issue with the ESP32 Arduino core or hardware and you'll have to report it to the ESP32 Arduino core maintainers or the board manufacturer.

morganrallen:
I can confirm with this code that it is not receiving an interrupt on the pin labeled 26.

morganrallen:
Calling `pinMode(26, INPUT);` before makes it work, but `onReceive` still is not.

sandeepmistry:
@morganrallen can you please try to add `pinMode(_dio0, INPUT);` here: https://github.com/sandeepmistry/arduino-LoRa/blob/master/src/LoRa.cpp#L281

If that works, please submit a pull request for the change, excellent debugging so far :)

morganrallen:
Hah, I was just trying exactly that. No such luck.

morganrallen:
[...]

PR #99, opened 2018-02-02
ensure pin DIO0 is set to input

morganrallen:
Thanks to @BlueCoal69 for suggesting it does work and making me think I missed something. Because.... I missed something.

While I was getting the pin to `INPUT` correctly, I failed to call `LoRa.receive`


morganrallen:
Fixes #83

sandeepmistry:
@morganrallen amazing, thanks!

Deltoz:
on Arduino MKRWAN1300 This line pinMode(_dio0, INPUT); blocking all.
I can't call the onReceive function.

PR #103, opened 2018-02-13
Make SPI port configurable at runtime

facchinm:


morganrallen:
This is great, #104 asks about using multiple radios, this could assist with that. I'll try to give this a test this week.

src/LoRa.cpp
```diff
@@ -50,6 +50,8 @@
 
 #define MAX_PKT_LENGTH           255
 
+static SPIClass& SPIPORT = SPI;
```
Rotzbua: Why not a member variable?

morganrallen: Agreed, in order to make this work with multiple SX1276 devices at the same time, this would need to be part of the class.

Rotzbua: @morganrallen For multiple devices are not multiple spi bus required. So 1 spi will work with 4-7 devices.
pic

morganrallen: Oh right, I even commented that on #104

wero1414: The important part is the interrupt of each device @morganrallen  are u working on this now?

morganrallen: I'm not actively working on this yet, I hope to get time tomorrow or Saturday

sandeepmistry:
@wero1414 @morganrallen I think a `LoRa.setSPI(...)` API would be a better fit rather than adding parameters to `begin` this was discussed with @facchinm. It's not specific to multiple radios. @facchinm's main goal was to get the MKR WAN "dumb" radio example working with this lib: https://github.com/arduino-libraries/MKRWAN/blob/master/examples/DumbModemLoraSender/DumbModemLoraSender.ino

Let me comment on #104 regarding multiple radios.

Issue #80, opened 2017-12-20
Add function to correct frequency


dontsovcmc:
We can correct frequency, that can go away little bit depence of temperature end etc.

4.1.5. Frequency Error Indication of http://www.semtech.com/images/datasheet/sx1276.pdf

I see only one code about this feature: https://electronix.ru/forum/index.php?showtopic=140311&hl=lora
But I can't understand some variables and not test it.

LoRaTracker:
Frequency correction for received packets might be desirable if you using bandwidths of 41.7khz and lower, and you can mitigate the effects somewhat by calibrating each module. 

However, whilst you can correct for frequency drift once you have established packet reception, if the RX and TX ends are outside the capture range (25% of bandwidth) packet reception will not start.

dontsovcmc:
Transmitter can't be outdoor and receiver indoor.  It will be big temperature difference in winter.

PR #111, opened 2018-02-24
Add getting the frequency error of a packet

torntrousers:
The LoRa chips frequency can drift over time and with things like the temperature, so a sender and receiver although set to the same frequency might not be on the same frequency. If they are too far apart then packets wont get received. This adds the ability to get the frequency error of a received packet so that you can then adjust the receiver frequency to more closely match that of the sender.

Code copied from my code here that I'd like to move from using the RadioHead LoRa library to use your one: https://github.com/HarringayMakerSpace/HabitatLoraGateway

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -25,7 +25,9 @@
 #define REG_PREAMBLE_LSB         0x21
 #define REG_PAYLOAD_LENGTH       0x22
 #define REG_MODEM_CONFIG_3       0x26
-#define REG_FREQ_ERROR           0x28
+#define REG_FREQ_ERROR_MSB       0x28
+#define REG_FREQ_ERROR_MID       0x29
+#define REG_FREQ_ERROR_LSB       0x2a
 #define REG_RSSI_WIDEBAND        0x2c
 #define REG_DETECTION_OPTIMIZE   0x31
 #define REG_DETECTION_THRESHOLD  0x37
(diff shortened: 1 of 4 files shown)


torntrousers:
Oh, I see someone else has already raised an issue about this: https://github.com/sandeepmistry/arduino-LoRa/issues/80

src/LoRa.cpp
```diff
@@ -211,6 +212,25 @@ float LoRaClass::packetSnr()
   return ((int8_t)readRegister(REG_PKT_SNR_VALUE)) * 0.25;
 }
 
+double LoRaClass::packetFrequencyError()
```
sandeepmistry: This looks good, any reason to use `double` over `float` here?

src/LoRa.cpp
```diff
@@ -365,6 +385,23 @@ void LoRaClass::setSpreadingFactor(int sf)
   writeRegister(REG_MODEM_CONFIG_2, (readRegister(REG_MODEM_CONFIG_2) & 0x0f) | ((sf << 4) & 0xf0));
 }
 
+long LoRaClass::getSignalBandwidth()
```
sandeepmistry: I'd prefer to leave this API out for now, unless we decide to have getters for all other parameters.

src/LoRa.cpp
```diff
@@ -211,6 +212,25 @@ float LoRaClass::packetSnr()
...
+  fe <<= 8L;
+  fe += (int32_t)readRegister(REG_FREQ_ERROR+1);
+  fe <<= 8L;
+  fe += (int32_t)readRegister(REG_FREQ_ERROR+2);
+  
+  if (readRegister(REG_FREQ_ERROR) & 8) {
+    fe = fe - 524288;
+  }
+
+  double freqError = (double)fe;
```
sandeepmistry: Are these formula's in the data sheet somewhere?

src/LoRa.cpp
torntrousers: Or does it even need the fraction part and perhaps should just be a ```long``` like ```setFrequency``` uses?

src/LoRa.cpp
torntrousers: Ok fair enough. Getters for the other settings could be good, but something for a separate pull.

src/LoRa.cpp
torntrousers: In the data sheet, section 4.1.5 "Frequency Error Indication"
The calculation code in https://github.com/sandeepmistry/arduino-LoRa/pull/113 looks a bit clearer. I wonder if that approach to be updating the frequency is better than just returning the frequency error value like this does?

torntrousers:
I've updated this as requested and merged in the error calculation code from  https://github.com/sandeepmistry/arduino-LoRa/pull/113. That code seemed better written and I tried them both at the same time and they both get the same result. I'd prefer to leave the adjusting of the frequency from the error to the user instead of the automatic approach used in compensateFrequencyOffset. Pull 113 has a couple of other things that would be good to do too in seperate pull requests - the fixing of the type of _frequency from int to long, and add a new public method to set the REG_PPM_CORRECTION that was happening in compensateFrequencyOffset

torntrousers:
Any comments on that last update?

sandeepmistry:
@morganrallen @universam1 @wero1414 any feedback on this before I merge?

sandeepmistry:
Note: the API.md file will also have to be updated for the new API.

halukmy:
thx

PR #113, opened 2018-02-25
Adding function for compensating frequency and ppm offset

szotsaki:
Call this function once after a successful package reception and
the centre frequency will be modified with the error indicated by
the LoRa chip. Along with this, PPM drift compensation also occurs.

The function returns the amount of the compensation needed.

Two more bug fixes:
- _frequency could be zero, now set to its default
- _frequency was stored only in `int` (aka `int16_t`) which caused that
  the stored value could be maximum 32767 Hz, not e.g. 930 000 000 Hz.
  - Because of this `packetRssi()` possible didn't work properly.

Please, independently check the validity of my calculation.

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -25,6 +25,10 @@
 #define REG_PREAMBLE_LSB         0x21
 #define REG_PAYLOAD_LENGTH       0x22
 #define REG_MODEM_CONFIG_3       0x26
+#define REG_PPM_CORRECTION       0x27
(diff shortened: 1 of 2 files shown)


Yury-MonZon:
I've tried this code and after the frequency adjustment, I can't receive the data anymore.

szotsaki:
Since this very pull request was basically merged in in #111 and in #119 by torntrousers (and they were tested), please, double check whether you have an other mistake in your code.

PR #142, opened 2018-04-23
Changed private to protected to enable more flexible

jappavoo:
Changed private to protected to enable more extension through inheritance.

Hi,

In my scenario I would like to use your class as a parent where the subclass can add methods that use the base methods for reading and writing the hardware registers.  

Jonathan.

### DIFF
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -66,7 +66,7 @@ public:
 
   void dumpRegisters(Stream& out);
 
-private:
+protected:
   void explicitHeaderMode();
   void implicitHeaderMode();
 


sandeepmistry:
Hi @jappavoo,

Thank you for submitting this pull request.

At this time I'm not too keen on open things up. Could you please expand on your use case? Thanks.

jappavoo:
Hi @sandeepmistry,

Thanks for your reply.  Sorry for the delay.  In my scenario I want to add new methods and reuse methods of your class.  Rather than modifying your code with things that are unstable or specific to my project I am using inheritance to create a subclass with all my changes.  However, for my subclass to have access to all the members and methods of your class I need them to be protected rather than private.  In this way I can access and reuse all your members and methods to prototype and add custom functionality that builds on your base functionality without messing around with your code.  

As an example the following code uses inheritance to extend the base class with functionality to 
read and dump states for my particular use case.  Of course the value is that my subclass also provides the public methods of your class as well as the additional methods of my subclass.

namespace MyProject {
class LoRaClass : public ::LoRaClass {
   unsigned char read_reg(uint8_t addr) { return readRegister(addr); }
#ifdef LORA_STATS
    uint16_t numValidHdrsRx_;
    uint16_t numValidPktsRx_;
    uint8_t  numPayLoadBytesRx_;
    uint8_t  irqFlags_;
    
    uint8_t getIRQFlags() {
      return read_reg(REG_LR_IRQFLAGS);
    }
    
    uint8_t getPayLoadBytesRx() {
      return read_reg(REG_LR_RXNBBYTES);
    }
    
    uint16_t getNumValidPktsRx() {
      uint16_t val;
      val = (read_reg(REG_LR_RXPACKETCNTVALUEMSB) << 8) |
  	    (read_reg(REG_LR_RXPACKETCNTVALUELSB)     );  
      return val;
    }
public:
   void dumpStats(Stream &s) {
      s.println("\tirqFlags: RXTMTOUT:" +
		String((irqFlags_ & RFLR_IRQFLAGS_RXTIMEOUT)!=0)  +
		" RXDONE:" +
		String((irqFlags_ & RFLR_IRQFLAGS_RXDONE)!=0) +
		" PLCRCERR:" +
		String((irqFlags_ & RFLR_IRQFLAGS_PAYLOADCRCERROR)!=0) +
		" VALIDHDR:" +
		String((irqFlags_ & RFLR_IRQFLAGS_VALIDHEADER)!=0) +
		" TXDONE:" +
		String((irqFlags_ & RFLR_IRQFLAGS_TXDONE)!=0) +
		" CADDONE:" +
		String((irqFlags_ & RFLR_IRQFLAGS_CADDONE)!=0) +
		" FHSSCHGCH:" +
		String((irqFlags_ & RFLR_IRQFLAGS_FHSSCHANGEDCHANNEL)!=0) +
		" CADDET:" +
		String((irqFlags_ & RFLR_IRQFLAGS_CADDETECTED)!=0));
      s.println("\tnumValidHdrsRx:" + String(numValidHdrsRx_) + 
              " numValidPktsRx:" + String(numValidPktsRx_) +
		" numPayLoadBytesRx:" + String(numPayLoadBytesRx_));  
    }

    bool updateStats() {
      bool chg = false;
      uint16_t irqFlgs = getIRQFlags();
      uint16_t hdrCnt  = getNumValidHdrRx();
      uint16_t pktCnt  = getNumValidPktsRx();
      uint8_t  bytes   = getPayLoadBytesRx();
      
      if (irqFlgs != irqFlags_)          {
	// dont't payattention to changes in RXTIMEOUTS output is too noisy
	if ((irqFlags_ ^ irqFlgs) != RFLR_IRQFLAGS_RXTIMEOUT) chg = true;
	irqFlags_=irqFlgs;
      }
      if (hdrCnt  != numValidHdrsRx_)    { numValidHdrsRx_=hdrCnt;   chg=true; }
      if (pktCnt  != numValidPktsRx_)    { numValidPktsRx_=pktCnt;   chg=true; }
      if (bytes   != numPayLoadBytesRx_) { numPayLoadBytesRx_=bytes; chg=true; }
      
      return chg;      
    }
    static const int MAX_PKT_SIZE = 255;
  } theLoRa;

}

sandeepmistry:
@jappavoo my preference is to hold off on this, the internal API's of the library may change in the future. So, I'm not keen on making them protected at this time.

jappavoo:
Makes sense.  Thanks.

sandeepmistry:
Let's close it for now @morganrallen.

PR #150, opened 2018-05-28
Better random number generation from RSSI

unprovable:
Hello,
The API call LoRa.random() gave users really bad random numbers. They need some level of whitening in order to be useful, so I've added some rather dirty (sorry!) code to implement a von Neumann Extractor. It's normally implemented in three functions, but I didn't want to bloat the class - feel free to edit it to make it more stylistic if you like :) See here for more information: https://en.wikipedia.org/wiki/Randomness_extractor#Von_Neumann_extractor

I tested the code that was originally there, and the random numbers were disappointing - I can send samples if you would like. This code will generate better random numbers. I also added a warning that these are not cryptographically secure, as I haven't proven that they are otherwise yet. :P

The speed is slightly slower, but very fast in the grander scheme of things (other entropy pools take a long time to generate entropy). I'm happy to answer any questions. 

Many thanks! M.

### DIFF
--- a/API.md
+++ b/API.md
@@ -343,7 +343,7 @@ Returns RSSI value as a byte.
 
 ### Random
 
-Generate a random byte, based on the Wideband RSSI measurement run through a von Neumann Extractor.
+Generate a random byte, based on the Wideband RSSI measurement run through a von Neumann Extractor. **NB** - these are not cryptographically secure random numbers! Use with caution!
 
 ```
 byte b = LoRa.random();
--- a/API.md
+++ b/API.md
@@ -331,9 +331,19 @@ LoRa.disableCrc();
 
 ## Other functions
 
+### Get RSSI
+
+Gets the RSSI value from the radio. 
+
+```
+byte b = LoRa.getRSSI();
+```
+
+Returns RSSI value as a byte.
+
 ### Random
 
-Generate a random byte, based on the Wideband RSSI measurement.
+Generate a random byte, based on the Wideband RSSI measurement run through a von Neumann Extractor.
 
 ```
 byte b = LoRa.random();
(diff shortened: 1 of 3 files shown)


sergio303:
FYI, here is Semtech implementation on SX1276 driver:

```
uint32_t SX1276Random( void )
{
    uint8_t i;
    uint32_t rnd = 0;

    /*
     * Radio setup for random number generation
     */
    // Set LoRa modem ON
    SX1276SetModem( MODEM_LORA );

    // Disable LoRa modem interrupts
    SX1276Write( REG_LR_IRQFLAGSMASK, RFLR_IRQFLAGS_RXTIMEOUT |
                  RFLR_IRQFLAGS_RXDONE |
                  RFLR_IRQFLAGS_PAYLOADCRCERROR |
                  RFLR_IRQFLAGS_VALIDHEADER |
                  RFLR_IRQFLAGS_TXDONE |
                  RFLR_IRQFLAGS_CADDONE |
                  RFLR_IRQFLAGS_FHSSCHANGEDCHANNEL |
                  RFLR_IRQFLAGS_CADDETECTED );

    // Set radio in continuous reception
    SX1276SetOpMode( RF_OPMODE_RECEIVER );

    for( i = 0; i < 32; i++ )
    {
        DelayMs( 1 );
        // Unfiltered RSSI value reading. Only takes the LSB value
        rnd |= ( ( uint32_t )SX1276Read( REG_LR_RSSIWIDEBAND ) & 0x01 ) << i;
    }

    SX1276SetSleep( );

    return rnd;
}
```

unprovable:
Aye - the extractor just helps to normalise the probabilities (P(0)=P(1)=0.5 under this scheme of output). It's not critical, but it's recommended to do some sort of whitening of this kind. :) Many thanks! M.

sandeepmistry:
Any updates on the hanging?

ricaun:
I'm messing with random this weekend and I found https://gist.github.com/endolith/2568571
I have tested some codes and get great results using left rotate and XOR.
Check https://github.com/sandeepmistry/arduino-LoRa/pull/188

wero1414:
There is another PR #395 where is added the seed necessary to have better random numbers

PR #153, opened 2018-05-31
sx127x 20dBm support 2

dontsovcmc:
https://github.com/sandeepmistry/arduino-LoRa/pull/79#issuecomment-386827531

### DIFF
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -54,6 +54,8 @@ public:
   void setSyncWord(int sw);
   void enableCrc();
   void disableCrc();
+  // Over Current Protection control (Semtech SX1276/77/78/79 5.4.4.)
+  void setOCP_sx127x(uint8_t mA);
 
   // deprecated
   void crc() { enableCrc(); }
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -10,6 +10,7 @@
 #define REG_FRF_MID              0x07
 #define REG_FRF_LSB              0x08
 #define REG_PA_CONFIG            0x09
+#define REG_OCP                  0x0b
 #define REG_LNA                  0x0c
 #define REG_FIFO_ADDR_PTR        0x0d
 #define REG_FIFO_TX_BASE_ADDR    0x0e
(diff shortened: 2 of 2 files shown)


Yury-MonZon:
This is working pretty good.

I've checked it with [this power meter](https://www.immersionrc.com/fpv-products/rfpowermeter/):

power setting 17 = 36mW (50mW) 
power setting 20 = 64mW (100mW)

It is a bit lower than theory (in parentheses) because the tx pulse is so short.  

When would this be merged to the master? @sandeepmistry

src/LoRa.h
```diff
@@ -54,6 +54,8 @@ class LoRaClass : public Stream {
   void setSyncWord(int sw);
   void enableCrc();
   void disableCrc();
+  // Over Current Protection control (Semtech SX1276/77/78/79 5.4.4.)
+  void setOCP_sx127x(uint8_t mA);
```
sandeepmistry: I don't think this needs to be a public API anymore. Maybe the name can also be changed to `setOCP` as well?

dontsovcmc:
Check, please =)

dontsovcmc:
who needs to get over current protection value?! =)

danak6jq:
Note that Semtech specifies a maximum duty cycle of 1% at +20dBm.

nevercast:
After speaking with Semtech they seem to think that as long as local regulations have no issue with your duty cycle, transmitting long duration payloads at +20dBm will be fine. This is obviously some mixed information. I asked if I can send 30 second long payloads at +20dBm and they said that will be fine.

danak6jq:
I am a little surprised if regulatory limits made it into the physical limits part of the Semtech datasheet, but I can understand it happening. If 30 seconds transmit is OK at +20dBm, that's 1% of 5 minutes. I don't recall off the top of my head what  the Part 15 conditions are for 1% but I believe this is well within the legal limits. Thank you for following-up with Semtech.

nevercast:
Regulatory limits in a Datasheet is certainly odd to me, in my country, at 900MHz there is no Duty Cycle requirement provided that the signal is Digitally Modulated. So the note about 1% really does not apply to me; unless it is for power dissipation concerns, such as transmitting continuously in high ambient temperature? 

I didn't get any more information from Semtech other than them saying that 30 second transmissions at +20dBm should not damage the device. So that's what I'll do :)

Issue #163, opened 2018-06-20
Sync Word defaults to 0x12


giampiero7:
Hi,
I was trying this library on an Arduino MKR WAN 1300 to communicate with a Pycom board.
It wouldn't work until I explicitly called `LoRa.setSyncWord(0x34);`
The default sync word seems to be 0x12, as opposed to what is stated in the documentation (0x34).
Cheers

morganrallen:
Unless I'm reading the datasheet wrong, `0x34` is the `LoRaWAN` sync word.

From page 115
**Reg:*** `RegSyncWord (0x39)`
**Bits:** `0-7`
**Name:** `SyncWord`
**Mode:** `rw`
**Reset:** `0x12`
**Description:** `LoRa Sync. Word Value 0x34 is reserved for LoRaWAN networks`

giampiero7:
Yes, but the documentation of this library (https://github.com/sandeepmistry/arduino-LoRa/blob/master/API.md#sync-word) says the default sync word is 0x34

morganrallen:
PR #175

sandeepmistry:
Closed via #175.

PR #175, opened 2018-07-30
Fix incorrect sync word in API.md

morganrallen:
API currently states the default sync word is 0x34 when in fact it's 0x12. 0x34 is the sync word when using LoRaWAN


sandeepmistry:
Closes #163.

PR #179, opened 2018-08-04
Add function to invert IQ signal register

ricaun:
InvertIQ allows differentiating sender and receiver message.

Usually, Gateways read messages with InvertIQ off and send messages with InvertIQ on, Nodes read messages with InvertIQ on and send messages with InvertIQ off.

This way a Gateway only reads messages from Nodes and never reads messages from other Gateway, and Node never reads messages from other Node.

### DIFF
--- a/examples/LoRaSimpleGateway/LoRaSimpleGateway.ino
+++ b/examples/LoRaSimpleGateway/LoRaSimpleGateway.ino
@@ -34,7 +34,7 @@ const int resetPin = 9;        // LoRa radio reset
 const int irqPin = 2;          // change for your board; must be a hardware interrupt pin
 
 void setup() {
-  Serial.begin(115200);                   // initialize serial
+  Serial.begin(9600);                   // initialize serial
   while (!Serial);
 
   LoRa.setPins(csPin, resetPin, irqPin);
--- a/examples/LoRaSimpleGateway/LoRaSimpleGateway.ino
+++ b/examples/LoRaSimpleGateway/LoRaSimpleGateway.ino
@@ -4,16 +4,16 @@
   This code uses InvertIQ function to create a simple Gateway/Node logic.
 
   Gateway - Sends messages with enableInvertIQ()
-          - Receives messeges with disableInvertIQ()
+          - Receives messages with disableInvertIQ()
 
   Node    - Sends messages with disableInvertIQ()
-          - Receives messeges with enableInvertIQ()
+          - Receives messages with enableInvertIQ()
 
-  With this arrangement a Gateway never receive messagem from another Gateway 
+  With this arrangement a Gateway never receive messages from another Gateway
   and a Node never receive message from another Node.
   Only Gateway to Node and vice versa.
 
-  This code receives messagens and sends a message every second.
+  This code receives messages and sends a message every second.
 
   InvertIQ function basically invert the LoRa I and Q signals.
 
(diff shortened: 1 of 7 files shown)


morganrallen:
Can you document the function in `API.md`?

ricaun:
I forgot to add on `API.md`, now is done

morganrallen:
Great! An example would also be awesome, not sure how to best demonstrate this though. Maybe have two devices what roles as the "gateway"? Open to any ideas but think there should be something in the examples folder for this.

src/LoRa.h
```diff
@@ -66,7 +66,8 @@ class LoRaClass : public Stream {
   void setSyncWord(int sw);
   void enableCrc();
   void disableCrc();
-
+  void invertIQ(boolean invert);
```
sandeepmistry: What do you think about making this:

* `enableInvertIQ()` and `disableInvertIQ()` to make it like the CRC API's above?

another option is `invertIQ()` and `noInvertIQ()`.

ricaun: Yeap, make sense InvertIQ be like CRC API's function.

examples/LoRaSimpleGateway/LoRaSimpleGateway.ino
```diff
@@ -0,0 +1,113 @@
...
+#include <LoRa.h>
+
+const long frequency = 915E6;  // LoRa Frequency
+
+const int csPin = 10;          // LoRa radio chip select
+const int resetPin = 9;        // LoRa radio reset
+const int irqPin = 2;          // change for your board; must be a hardware interrupt pin
+
+void setup() {
+  Serial.begin(115200);                   // initialize serial
```
sandeepmistry: Let's have the default baud at 9600 like the other examples, as it's the default baud rate of the Arduino IDE.

examples/LoRaSimpleNode/LoRaSimpleNode.ino
```diff
@@ -0,0 +1,113 @@
...
+#include <LoRa.h>
+
+const long frequency = 915E6;  // LoRa Frequency
+
+const int csPin = 10;          // LoRa radio chip select
+const int resetPin = 9;        // LoRa radio reset
+const int irqPin = 2;          // change for your board; must be a hardware interrupt pin
+
+void setup() {
+  Serial.begin(115200);                   // initialize serial
```
sandeepmistry: Similar comment as above.

src/LoRa.h
```diff
@@ -66,11 +66,16 @@ class LoRaClass : public Stream {
...
+  void enableInvertIQ();
+  void disableInvertIQ();
+  
   void setOCP(uint8_t mA); // Over Current Protection control
 
   // deprecated
   void crc() { enableCrc(); }
   void noCrc() { disableCrc(); }
+  void invertIQ() { enableInvertIQ(); }
+  void noInvertIQ() { disableInvertIQ(); }
```
sandeepmistry: I'd say we remove `invertIQ()` and `noInvertIQ()` instead of adding deprecated methods.

sandeepmistry:
@ricaun thanks!

src/LoRa.h
GOWSITH: sir is it possible to communicate  RN2483 with sx1276 we tried but data not receiving in rn2483 but sx1276 receives data from rn2483 is it because of serial and spi communication

PR #188, opened 2018-09-07
Callback onTxDone and Better Random

ricaun:
It's possible to use dio0 interrupt to create a callback when transmission finish.
And a better random using the rotate left and XOR.


ricaun:
> 
> 
> Can you split this into two separate pull requests?

No problem, I create two new pull requests!

PR #189, opened 2018-09-14
isTransmitting() public

giampiero7:
Hi,
I just updated the `isTransmitting()` method to be public.
When transmitting using the blocking functions added in #62 this function is necessary externally to calculate transmission time (to respect duty cycles).
BTW, it would be very useful to have a lower level function to get actual transmission time of the last sent packet.


SanValens (approved):

giampiero7:
Can this be merged please?

PR #213, opened 2018-12-22
Improve Random Function

ricaun:
Improve random function using the rotate left and XOR.
Reference: https://gist.github.com/endolith/2568571
I create an example hows generate a random seed using LoRa.random().


morganrallen:
Closing this as there are too many unrelated commits for a single PR

PR #214, opened 2018-12-22
Callback onTxDone

ricaun:
It's possible to use dio0 interrupt to create a callback when transmission finish and when the transceiver receives some data. 
The register REG_DIO_MAPPING_1 is responsible for this interrupts.


morganrallen:
Hi @ricaun ! I'd love to merge this but there are like 20 unrelated commits in the PR. Can you resubmit with only the commits directly related to this feature?

ricaun:
Yep is really mess up, I created a new one! #320

Issue #254, opened 2019-06-05
Crash: ISR not in IRAM when 


sbrl:
When I try to use continuous receive mode with a Wemos D1 R2, I get the following crash:

```
Abort called

>>>stack>>>

ctx: cont
sp: 3ffffda0 end: 3fffffc0 offset: 01b0
3fffff50:  000000c0 00000000 3ffee290 4020154f  
3fffff60:  007a1200 00000001 0000000f 3ffee340  
3fffff70:  3fffdad0 3ffee294 3ffee294 4010044a  
3fffff80:  3fffdad0 3ffee294 3ffee294 40201834  
3fffff90:  3fffdad0 3ffee294 3ffee2e8 4020108d  
3fffffa0:  feefeffe 00000000 3ffee310 40202448  
3fffffb0:  feefeffe feefeffe 3ffe84f4 401007c5  
<<<stack<<<
?܆⸮⸮D⸮N⸮LoRa Receiver
ISR not in IRAM!
```

Here's a link to the sketch I'm using: https://privatebin.net/?f92cda1dbecb73ee#J6+bp9JBfSdUhJ5S7X/u1YsOrPxE4rQddfJgH736wpg=

torntrousers:
Which release of the Arduino ESP8266 core are you using? I think in the latest core releases all the ISR's must use ICACHE_RAM_ATTR so we probably need to update this LoRa library for that. 
Can you try add ICACHE_RAM_ATTR to the start of [this line](https://github.com/sandeepmistry/arduino-LoRa/blob/master/src/LoRa.cpp#L681) to see if it fixes it?

sbrl:
Hey, thanks so much for the lightning fast reply @torntrousers! :zap:

Changing that line to this does indeed fix the issue:

```c++
ICACHE_RAM_ATTR void LoRaClass::onDio0Rise()
```

However, this change causes compilation to fail for the Arduino Uno (I also have 2 of these - 1 with a Dragino LoRa Shield that I'm borring from University until ~August, and another which has an RFM95 connected via level shifter).

Looking in the boards manager reveals that I'm using v2.5.2 of the esp8266 package, if that's the one you're talking about?

torntrousers:
I guess that should be surrounded by a conditional, something like - ```#ifdef ESP32 || ESP8266```, https://community.platformio.org/t/how-change-define-device-platform/4251

morganrallen:
[...]

PR #257, opened 2019-06-08
Fix Crash: ISR not in IRAM when #254

torntrousers:
Fix for https://github.com/sandeepmistry/arduino-LoRa/issues/254


sbrl:
Yep, this works :rocket:

sbrl:
:tada:

torntrousers:
FYI, see the comment [here](https://github.com/esp8266/Arduino/issues/6127#issuecomment-501687860). I did wonder about all the code thats in handleDio0Rise. The update in this PR does seem ok to fix the immediate 'ISR not in IRAM' crash but the approach to how the DIO interrupt is handled should probably be revisited.

Issue #268, opened 2019-07-11
Please label it properly


Garfius:
Once long tested the code, and seen so many issues, and understanding the hardships of the general prupose. I understand this code has no possible professional application.

Therefore, i sincerely ask to modify readme.md to show a warning message regarding a demonstration prupose.

Thanks.

torntrousers:
It works well for me. Would you list out the issues you've seen?

Garfius:
The most important:

Can't work nore than a few hours.

Interrupt driven receive method is not reliable.

mchacher:
Please read following issues. I did the patch on my side.
    // https://github.com/sandeepmistry/arduino-LoRa/issues/218
    // https://github.com/sandeepmistry/arduino-LoRa/issues/222
Since the callback is done under interupt, limit as much as possible the code in the ISR.

torntrousers:
If you've got a patch for that issue would you open a PR or show the diff?

mchacher:
Just comment line 654 as explained here.
https://github.com/mchacher/LoRa2MQTT/commit/f03fc0bf020fc4ee5898b92cff39429a2ec926a0

mchacher:
@torntrousers , just comment line :
_writeRegister(REG_FIFO_ADDR_PTR, 0);_
at end of the function void LoRaClass::handleDio0Rise()

Issue #218, opened 2019-01-17
Any reason for resetting REG_FIFO_ADDR_PTR in LoRaClass::handleDio0Rise() ?


iw2lsi:
Hi all,

 just a doubt..  what's the reason of resetting REG_FIFO_ADDR_PTR register at the end of the handleDio0Rise() function ?

I cannot find anything pointing in this direction in the datasheets... and it seems to me dangerous in case another packet is incoming while we are executing the _onReceive()... 

If this is fine... shouldn't we pospone the reset of the IRQ flags after the  line:

writeRegister(REG_FIFO_ADDR_PTR, 0) ?

    Best Regards

            Giampaolo

keithr0:
This caused me problems with an interrupt driven receive application giving incorrect packet contents until that statement was removed. REG_FIFO_ADDR_PTR should point to the start of the current packet as is done a couple of statements back with 
writeRegister(REG_FIFO_ADDR_PTR, readRegister(REG_FIFO_RX_CURRENT_ADDR));
Resetting it to zero does not appear to make any sense.

notascooby:
I have same issue with interrupt driven receive. Master sends a command, slave acknowledges, then acknowledges a second time 4 seconds after first acknowledgement. The LoRa buffer includes both the first acknowledgement and part of the second. ESP32 has to be reset before OnReceive works again.
Deleting REG_FIFO_ADDR_PTR at line 584 in LoRa.cpp (library version 0.5.0) solves problem.
Haven't seen any problems after 1 day continuous use, with messages of around 15 characters.

iw2lsi:
That's fine...

  we are also working without  writeRegister(REG_FIFO_ADDR_PTR, 0) at line 583 and it seems to work... 

Sandeep what do you think ?

morganrallen:
Hi just wanted to say I've been following this issue. I have a strong recollection there is a specific reason the reset occurs early (I do the same in esp32-lora). Can someone provide a concrete test case I can work with later this week?

notascooby:
Thanks iw2lsi, will do and if I see any issues message them here.

sandeepmistry:
@iw2lsi the fastest way to move on this would be to open a pull request with the suggested, and also provide two sketches (sender and receiver) that be used to reproduce the issue as @morganrallen suggested above.

notascooby:
Have written two sketch’s yesterday but need to test to check they exhibit problem.
[...]

PR #269, opened 2019-07-12
Don't reset REG_FIFO_ADDR_PTR

torntrousers:
As per the discussion in https://github.com/sandeepmistry/arduino-LoRa/issues/268 and https://github.com/sandeepmistry/arduino-LoRa/issues/218

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -652,9 +652,6 @@ void LoRaClass::handleDio0Rise()
     if (_onReceive) {
       _onReceive(packetLength);
     }
-
-    // reset FIFO address
-    writeRegister(REG_FIFO_ADDR_PTR, 0);
   }
 }
 


morganrallen:
After digging into the datasheet I cannot find a compelling reason resetting REG_FIFO_ADDR_PTR should be done.

mchacher:
I do confirm.
could be good as well to create a task that is calling the callback function instead of having this call in the ISR which is not safe at all.
Sharing below the code I have used on my side.

> /**
 * onDio0Rise
 * Handle Dio0 interrupt
 */
void IRAM_ATTR LoRaClass::onDio0Rise()
{
  portENTER_CRITICAL(&mutexDioRise);
  _dioRiseInterruptCounter++;
  portEXIT_CRITICAL(&mutexDioRise);
  //LoRa.handleDio0Rise();
}

/**
 * Manage RX / LoRa data received
 */
void LoRaClass::taskRxLoop(void *pvParameters)
{
  // configure dio0 interrupt
  vTaskDelay(1);
  pinMode(LoRa._dio0, INPUT);
  LoRa.writeRegister(REG_DIO_MAPPING_1, 0x00);
  attachInterrupt(digitalPinToInterrupt(LoRa._dio0), LoRaClass::onDio0Rise, RISING);
  while (true)
  {
    if (_dioRiseInterruptCounter > 0)
    {
      portENTER_CRITICAL(&mutexDioRise);
      _dioRiseInterruptCounter--;
      portEXIT_CRITICAL(&mutexDioRise);
      LoRa.handleDio0Rise();
    }
    // give the opportunity to the IDLE task to run, and so avoid the TaskWatchDog timer to trigger a reset
    vTaskDelay(10);
  }
}

morganrallen:
This isn't related to this pull request.

PR #280, opened 2019-08-08
ifdef -> if defined

saloid:
ifdef not supports || (or) condition


morganrallen:
Thanks for the PR, a variation of this has been pulled in.

PR #288, opened 2019-09-17
fix mid band threshold and add function to get current RSSI value

sergio303:
Mid band threshold used for RSSI calculation is wrong, causing incorrect RSSI values for frequencies on the HF band but below 868 MHz . Semtech reference code uses 525 MHz as threshold for RSSI calculation:
```
/*!
 * Constant values need to compute the RSSI value
 */
#define RSSI_OFFSET_LF                              -164.0
#define RSSI_OFFSET_HF                              -157.0
#define RF_MID_BAND_THRESH                          525000000
```

### DIFF
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -39,6 +39,8 @@ public:
   float packetSnr();
   long packetFrequencyError();
 
+  int rssi();
+
   // from Print
   virtual size_t write(uint8_t byte);
   virtual size_t write(const uint8_t *buffer, size_t size);
--- a/API.md
+++ b/API.md
@@ -169,7 +169,7 @@ The `onReceive` callback will be called when a packet is received.
 ### Packet RSSI
 
 ```arduino
-int pRssi = LoRa.packetRssi();
+int rssi = LoRa.packetRssi();
 ```
 
 Returns the averaged RSSI of the last received packet (dBm).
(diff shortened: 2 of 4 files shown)


torntrousers:
Watching with interest. Update this too? - https://github.com/sandeepmistry/arduino-LoRa/blob/master/API.md

morganrallen:
Yeah, please add the new rssi method to the API otherwise this looks good to me. @torntrousers or anyone else have thoughts? I'd read through the datasheet again the merge this if there are no concerns.

torntrousers:
LGTM

sergio303:
API updated. New rssi method can be used to watch channel activity.
I'm also looking at packet rssi calculation in Semtech datasheet and code, it looks like some adjustment is needed when packets are received below the noise floor (snr <0), will submit a separate PR when I sort things out...

IoTThinks:
@morganrallen @sandeepmistry I realized I reinvented the wheel today by doing the same code.
Could you help to review and to push this to your repository?

image

src/LoRa.h
```diff
@@ -39,6 +39,8 @@ class LoRaClass : public Stream {
   float packetSnr();
   long packetFrequencyError();
 
+  int rssi();
```
sandeepmistry, line 42: this needs to go in `keywords.txt`. I'll try to edit it now.

sandeepmistry (approved):
@sergio303 thanks!

sandeepmistry:
@IoTThinks thanks for nudging this PR :)

IoTThinks:
Later, I will test on the merge.

Thanks a lot for your effort.

PR #297, opened 2019-10-09
Add compatibility with MKRWAN1310

facchinm:
IRQ pin is now exported and can be used to avoid polling. However, the extra ASM instruction in "dumb mode" requires a slower SPI clock


facchinm:
Sure, will do!

PR #329, opened 2020-01-31
onError callback

artemen:
added a callback to notify user about a failed CRC packet.

### DIFF
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -58,6 +58,7 @@ public:
 #ifndef ARDUINO_SAMD_MKRWAN1300
   void onReceive(void(*callback)(int));
   void onTxDone(void(*callback)());
+  void onError(void(*callback)());
 
   void receive(int size = 0);
 #endif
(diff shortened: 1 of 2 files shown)


src/LoRa.h
```diff
@@ -58,6 +58,7 @@ class LoRaClass : public Stream {
 #ifndef ARDUINO_SAMD_MKRWAN1300
   void onReceive(void(*callback)(int));
   void onTxDone(void(*callback)());
+  void onError(void(*callback)());
```
sandeepmistry: I'm thinking this should be renamed to something more specific, for example: `onCrcError` or `onReceiveCrcError`. 

Thoughts?

artemen: Please name it in accordance with the rest of the repo. No objections.

morganrallen: Hi @artemen if you can address these minor changes I'll merge this PR and include in the next release, which I'd like to get to this week.

Thanks for this contribution, it will add some important insight to error conditions.

artemen:
sorry for the bit of mess. let me know if I should do a different pull request or this one will do, thanks!

szerwi:
Do you plan to merge this PR?

PR #331, opened 2020-02-02
Channel Activity Detection function and examples.

w-ockham:
Added Channel Activity Detection and its examples.

### DIFF
--- a/API.md
+++ b/API.md
@@ -40,7 +40,7 @@ To save further pins one could connect the reset pin of the MCU with reset pin o
 
 #### Pin dio0 interrupt callbacks
 
-The dio0 pin can be used for transmission finish callback and/or receiving callback, check `onTxDone` and `onReceive`.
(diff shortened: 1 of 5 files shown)


artemen:
hi, have you tested if this could be used to  "catch" a partial preamble, than switch into RX mode and read the packet?

w-ockham:
Not yet with the packet which has long preamble, but works fine with low rate mode. It captures the preamble at the beginning of the packet then transit to continuous receive mode.
Unfortunately, I found a mistake at the interrupt handler (CAD should be handled before the  PAYLOAD_CRC_ERROR body clause).  Withdraw this request and will send another one after running some tests on the long preamble packet.

IoTThinks:
HI all,
Will this CAD be considered stable to be in a new release?
I'm eager to test on it.
Thanks a lot. :D

PR #333, opened 2020-02-03
Add ability for receiver to check if CRC is present in payload

mcgurk:
enableCrc() affects only to sender. This modification adds CrcOnPayload()-function, which can be used to detect if CRC is present in payload.

### DIFF
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -73,6 +73,7 @@ public:
   void setSyncWord(int sw);
   void enableCrc();
   void disableCrc();
+  uint8_t CrcOnPayload();
   void enableInvertIQ();
   void disableInvertIQ();
   
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -20,6 +20,7 @@
 #define REG_RX_NB_BYTES          0x13
 #define REG_PKT_SNR_VALUE        0x19
 #define REG_PKT_RSSI_VALUE       0x1a
+#define REG_HOP_CHANNEL          0x1c
 #define REG_MODEM_CONFIG_1       0x1d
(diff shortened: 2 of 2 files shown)


src/LoRa.h
```diff
@@ -73,6 +73,7 @@ class LoRaClass : public Stream {
   void setSyncWord(int sw);
   void enableCrc();
   void disableCrc();
+  uint8_t crcOnPayload();
```
sandeepmistry: I'm thinking this could be renamed to something more consistent with: https://github.com/sandeepmistry/arduino-LoRa/blob/master/src/LoRa.h#L43-L46

Maybe: `packetHasCrc()`?

Also, for the return type type is `bool` a better option?

Thoughts?

IoTThinks: isCrcPresent?

Normally a bool variable may start with is...?

mcgurk: As far as I know, in Arduino IDE `bool` is same than `uint8_t`. I have habit to use uint8_t, but I have nothing against bool. bool is just little bit more abstract.

I think too that crcOnPayload is little problematic function name. It is accurate, but it doesn't indicate what it means to reader. Both your suggestions are better than "crcOnPayload".

morganrallen: Hi @mcgurk, if you can address these minor changes I'll merge this PR and include in the next release I'd like to get out this week. Thanks for this contribution and keen eyed reading of the datasheet.

mcgurk: Hmm.. This is my first pull request ever :D. What I have to do? Edit my branch? I try.

morganrallen: Yeah you just push additional commits to your branch and they will show up here.

PR #348, opened 2020-03-11
Version 0.7.1

ricaun:
I would like to force a new version and work with the sandeepmistry /arduino-LoRa master repository.

I have a [project LoRaNow](https://github.com/ricaun/loranow) that uses a forked repository of arduino-LoRa and with the `depends=` on the `library.properties` file I could force the Arduino IDE to download the original repository.

But I use the `OnTxDone` function and version 0.7.0 doesn't have.

Could you force a new version?

I will appreciate that!

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=LoRa
-version=0.7.0
+version=0.7.1
 author=Sandeep Mistry <sandeep.mistry@gmail.com>
 maintainer=Sandeep Mistry <sandeep.mistry@gmail.com>
 sentence=An Arduino library for sending and receiving data using LoRa radios.


morganrallen:
Yes! I've been testing a couple parts I'd like to include, I should able to get to this tonight.

morganrallen:
This has been merged but after playing with it a bit I don't think the example really conveys what this is useful for. You might have seen I increased the wait timeout to 5 seconds because I thought 1 wasn't even but even after tweaking spreading factor 5 wasn't. I'd like to see the example updated to actually wait for TXDone to be called, **then** wait 5 seconds and send again.

morganrallen:
Oh, also reporting the TX time would be a nice addition.

ricaun:
> 
> 
> This has been merged but after playing with it a bit I don't think the example really conveys what this is useful for. You might have seen I increased the wait timeout to 5 seconds because I thought 1 wasn't even but even after tweaking spreading factor 5 wasn't. I'd like to see the example updated to actually wait for TXDone to be called, **then** wait 5 seconds and send again.

Yes, 1 second could bee too aggressive, 5 seconds is a better call. 
In the example is sending every 5 seconds, we can change and do something like this...

```
#include <SPI.h>
#include <LoRa.h>

int counter = 0;
bool isBusy = false;
unsigned long currentMillis = 0;

void setup() {
  Serial.begin(9600);
  while (!Serial);

  Serial.println("LoRa Sender non-blocking Callback");

  if (!LoRa.begin(915E6)) {
    Serial.println("Starting LoRa failed!");
    while (1);
  }

  LoRa.onTxDone(onTxDone);
}

void loop() {
  if (!isBusy) { 
    Serial.print("Sending packet non-blocking: ");
    Serial.println(counter);

    // send in async / non-blocking mode
    LoRa.beginPacket();
    LoRa.print("hello ");
    LoRa.print(counter);
    LoRa.endPacket(true); // true = async / non-blocking mode
    currentMillis = millis();
    isBusy = true;
    counter++;
  }
  delay(5000);
}

void onTxDone() {
  Serial.print("TxDone ");
  Serial.println(millis()-currentMillis);
  isBusy = false;
}
```

I don't know, I think the idea of the `onTxDone` is not blocking the main loop, I don't know where to put the `delay(5000);`.

> 
> 
> Oh, also reporting the TX time would be a nice addition.

I add some `Serial.println(millis()-currentMillis);`, I think the example should be as simpler as possible.

morganrallen:
You could...
* mark the `start time` & `isBusy` in the loop just before sending
* then mark the `end time` & `isBusy` to `false` in `onTxDone`
* in the loop check `isBusy`
* once not busy, report the time (remember, Serial.print from an interrupt is bad)
* if `millis() - endTime < 5000 return`

Also, `isBusy` isn't very descriptive, how about `isSending`?

ricaun:
Looks nice, whats do you think?!

```
#include <SPI.h>
#include <LoRa.h>

int counter = 0;
bool isSending = false;
unsigned long time = 0;

void setup() {
  Serial.begin(9600);
  while (!Serial);

  Serial.println("LoRa Sender non-blocking Callback");

  if (!LoRa.begin(915E6)) {
    Serial.println("Starting LoRa failed!");
    while (1);
  }

  LoRa.onTxDone(onTxDone);
}

void loop() {
  if (millis() - time > 5000) {
    if (!isSending) {
      Serial.print("Sending packet non-blocking: ");
      Serial.println(counter);

      // send in async / non-blocking mode
      LoRa.beginPacket();
      LoRa.print("hello ");
      LoRa.print(counter);
      LoRa.endPacket(true); // true = async / non-blocking mode
      isSending = true;
      counter++;
    }
  }
}

void onTxDone() {
  Serial.println("TxDone ");
  time = millis();
  isSending = false;
}
```

IoTThinks:
You can check the airtime here.
https://www.loratools.nl/#/airtime

For me if I use SF7 and short messages (around 50 bytes). I can send messages at 500ms inverval.
So it's up to the air time.

BTW, 5s is good for a general case (Up to SF12, 125Khz and 100 byte message).

morganrallen:
That's not the point though, it shouldn't send at all before the last one is gone

IoTThinks:
Hi,
I can see and install v0.7.1 in Arduino IDE already.
Thanks all for the effort. :D

PR #358, opened 2020-04-12
Examples: replaced hardcoded value

Chornomorets:
- replaced hardcoded value with existed variable that contains same value;


dhorner:
The reason why it was Hardcoded is that it is a check to see if it was a Broadcast message (0xFF).  If you changed the destination to another address (other than broadcast) then the change breaks it the code.

It would be better to add:

`#define BROADCAST_ADDRESS 0xFF
`
and use that in the code instead of destination.

PR #389, opened 2020-08-12
Add simple function for measuring temperature

ladislavkrivy:


IoTThinks:
@ladislavkrivy How good is the builtin temperature?
I thought the temperature measure can only be done in FSK mode?

ladislavkrivy:
It's quite terrible but good enough for compensating crystal's temperature drift (on small crystals tends to be linear-ish). In my case just with linear compensation by `LoRa.setFrequency(F_UPLINK + 150 * LoRa.temperature());` before sending a packet I can easily get away with 20.8k bandwidth. 
Screenshot_20200813_180257

Yes, to measure the temperature it switches to FSK mode momentarily and restores the mode afterwards. 
It could be improved if there's a way to check wether the module is currently receiving a packet

IoTThinks:
Thanks a lot for your code.
Let me try next week.

It should be good for LoRa node in outdoor environment?

ladislavkrivy:
Cool, It'll be interesting to know whether the temperature characteristic is somewhat similar for those 32M crystals (was for mine but they are probably same batch) or actually it has wilder variance.

Keep in mind the reading is not calibrated so you'll probably need to use frequencyError for initial calibration. I have 3 outdoor nodes at different places and with temperature compensation the system works even when sun hits one of the nodes.

ned-kelly:
Any chance we can merge this one in?

IoTThinks:
Should be a good feature to merge in the main.

Issue #394, opened 2020-09-07
byte LoRaClass::random() Improper Handling of RegRssiWideband for TRNG


Kongduino:
LoRaClass::random() reads the register once and returns the byte.

```c
byte LoRaClass::random() {
  return readRegister(REG_RSSI_WIDEBAND);
}
```

This is wrong for two reasons:

1) LoRa chip must be set in a specific mode before that, you can't get random number at any given time. LoRa must be stopped and set to continuous mode first, and so on.

2) There should be 8 reads, keeping only the least significant bit (LSB), and build the output byte from that.

See Chapter 4 of [Random Number Generation for Cryptography](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/440000001NAw/7YN8ZamV70_xR.vPDAAAshm.0Wt4jmRX0nOKkOzQqiI).

> To generate an N bit random number, perform N read operation of the register RegRssiWideband (address 0x2c) and use the LSB of the fetched value. The value from RegRssiWideband is derived from a wideband (4MHz) signal strength at the receiver input and the LSB of this value constantly and randomly changes.

Example code from [LoRandom.h](https://github.com/Kongduino/BastWAN_LoRa_RNG/blob/master/LoRandom.h)

```C
#define RegOpMode 0x01
#define RegModemConfig1 0x1D
#define RegModemConfig2 0x1E
#define RegRssiWideband 0x2C

void setupLoRandom() {
  LoRa.writeRegister(RegOpMode, 0b10001101);
  // LoRa mode enable Receive Continuous mode
  LoRa.writeRegister(RegModemConfig1, 0b01110010);
  // ‘0111’ for 125kHz modulation Bandwidth
  // 4/5 error coding rate
  // Packets have up-front header
  LoRa.writeRegister(RegModemConfig2, 0b01110000);
  // ‘0111’ (SF7) = 6kbit/s
}

uint8_t getLoRandomByte() {
  uint8_t x = 0;
  for (uint8_t j = 0; j < 8; j++) {
    x += (LoRa.readRegister(RegRssiWideband) & 0b00000001);
    x = x << 1;
    delay(1);
  }
  return x;
}
```

You need to put the SX1276/8 in Receive Continuous mode, SF7, C/R 4/5, BW 125KHz (call to `setupLoRandom()`, and then read 8 bytes to get one random byte (call to `getLoRandomByte()`). After which you need to put back your LoRa chip to whatever settings you were using.

[...]

PR #395, opened 2020-09-07
Fix Random function 

sabas1080:
Hi everyone

I am doing this pull request based in the work of @Kongduino in the issue #394 

Welcome feedback

Thanks

### DIFF
--- a/examples/LoRandom/LoRandom.ino
+++ b/examples/LoRandom/LoRandom.ino
@@ -1,0 +1,88 @@
+/*
+  LoRandomSeed
+
+  Example demonstrating how to generate random numbers using LoRa.
+
+  created 7 Sep 2020
+  by Konduino
+  adapted by Andres Sabas
+  
+  SX1276 Register (address)     Register bit field (bit #)      Values    Note
+  RegOpMode (0x01)              LongRangeMode[7]                ‘1’       LoRa mode enable
+                                Mode[2:0]                       ‘101’     Receive Continuous mode
+  ------------------------------------------------------------------------------------------------------------------
+  RegModemConfig1 (0x1D)        Bw[7:4]                         ‘0111’    ‘0111’ for 125kHz modulation Bandwidth
+                                CodingRate[3:1]                 ‘001’     4/5 error coding rate
+                                ImplicitHeaderModeOn[0]         ‘0’       Packets have up-front header
+  ------------------------------------------------------------------------------------------------------------------
+  RegModemConfig2 (0x1E)        SpreadingFactor[7:4]            ‘0111’    ‘0111’ (SF7) = 6kbit/s
+
+  To generate an N bit random number, perform N read operation of the register RegRssiWideband (address 0x2c)
+  and use the LSB of the fetched value. The value from RegRssiWideband is derived from a wideband (4MHz) signal strength
+  at the receiver input and the LSB of this value constantly and randomly changes.
+*/
+
+#include <SPI.h>
+#include <LoRa.h>
+
+void hexDump(unsigned char *buf, uint16_t len) {
+  String s = "|", t = "| |";
+  Serial.println(F("  |.0 .1 .2 .3 .4 .5 .6 .7 .8 .9 .a .b .c .d .e .f |"));
+  Serial.println(F("  +------------------------------------------------+ +----------------+"));
+  for (uint16_t i = 0; i < len; i += 16) {
+    for (uint8_t j = 0; j < 16; j++) {
+      if (i + j >= len) {
+        s = s + "   "; t = t + " ";
+      } else {
+        char c = buf[i + j];
+        if (c < 16) s = s + "0";
+        s = s + String(c, HEX) + " ";
+        if (c < 32 || c > 127) t = t + ".";
+        else t = t + (char)c;
+      }
+    }
+    uint8_t index = i / 16;
+    Serial.print(index, HEX); Serial.write('.');
+    Serial.println(s + t + "|");
+    s = "|"; t = "| |";
+  }
+  Serial.println(F("  +------------------------------------------------+ +----------------+"));
+}
+
+void setup() {
+  Serial.begin(115200);
+  while(!Serial);
+  Serial.print(F("\n\n\n[SX1276] Initializing ... "));
+  LoRa.setPins(SS, RFM_RST, RFM_DIO0);
+  Serial.println("SS: " + String(SS));
+  Serial.println("RFM_RST: " + String(RFM_RST));
+  Serial.println("RFM_DIO0: " + String(RFM_DIO0));
+  Serial.println("RFM_SWITCH: " + String(RFM_SWITCH));
+  if (!LoRa.begin(868E6)) {
+    Serial.println("Starting LoRa failed!");
+    while (1);
+  }
+  Serial.print("Setting up LoRa ");
+  pinMode(RFM_SWITCH, OUTPUT);
+  digitalWrite(RFM_SWITCH, 1);
+  LoRa.setTxPower(20, PA_OUTPUT_PA_BOOST_PIN);
+  LoRa.setPreambleLength(8);
+  LoRa.LoRandomSeed();
+  Serial.println("[o]");
+  delay(1000);
+  Serial.println("End of setup\n\n");
+}
+
+void loop() {
+  unsigned char randomStock[256];
+  // We'll build a stock of random bytes for use in code
+  uint8_t randomIndex = 0;
+  uint16_t i;
+  for (i = 0; i < 256; i++) {
+    uint8_t x = LoRa.random();
+    randomStock[i] = x;
+  }
+  randomIndex = 0;
+  hexDump(randomStock, 256);
+  delay(2000);
+}
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -607,9 +607,21 @@ void LoRaClass::setOCP(uint8_t mA)
   writeRegister(REG_OCP, 0x20 | (0x1F & ocpTrim));
 }
 
+void LoRaClass::LoRandomSeed() {
+  writeRegister(REG_OP_MODE, 0b10001101);
+  writeRegister(REG_MODEM_CONFIG_1, 0b01110010);
+  writeRegister(REG_MODEM_CONFIG_2, 0b01110000);
+}
+
 byte LoRaClass::random()
 {
-  return readRegister(REG_RSSI_WIDEBAND);
+  uint8_t x = 0;
+  for (uint8_t j = 0; j < 8; j++) {
+    x += (readRegister(REG_RSSI_WIDEBAND) & 0b00000001);
+    x = x << 1;
+    delay(1);
+  }
+  return x;
 }
 
 void LoRaClass::setPins(int ss, int reset, int dio0)
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -81,7 +81,8 @@ public:
   // deprecated
   void crc() { enableCrc(); }
   void noCrc() { disableCrc(); }
-
+  
+  void LoRandomSeed();
   byte random();
 
   void setPins(int ss = LORA_DEFAULT_SS_PIN, int reset = LORA_DEFAULT_RESET_PIN, int dio0 = LORA_DEFAULT_DIO0_PIN);


examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,88 @@
...
+  Serial.println("SS: " + String(SS));
+  Serial.println("RFM_RST: " + String(RFM_RST));
+  Serial.println("RFM_DIO0: " + String(RFM_DIO0));
+  Serial.println("RFM_SWITCH: " + String(RFM_SWITCH));
+  if (!LoRa.begin(868E6)) {
+    Serial.println("Starting LoRa failed!");
+    while (1);
+  }
+  Serial.print("Setting up LoRa ");
+  pinMode(RFM_SWITCH, OUTPUT);
```
wero1414, line 66: This will not work for all boards, maybe we can add a macro for #ifdef SAMR34

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,88 @@
...
+  digitalWrite(RFM_SWITCH, 1);
```
wero1414, line 67: Same here

wero1414, line 67: This will not work for all boards, maybe we can add a macro for #ifdef SAMR34

src/LoRa.cpp
```diff
@@ -607,9 +607,21 @@ void LoRaClass::setOCP(uint8_t mA)
   writeRegister(REG_OCP, 0x20 | (0x1F & ocpTrim));
 }
 
+void LoRaClass::LoRandomSeed() {
+  writeRegister(REG_OP_MODE, 0b10001101);
+  writeRegister(REG_MODEM_CONFIG_1, 0b01110010);
+  writeRegister(REG_MODEM_CONFIG_2, 0b01110000);
```
wero1414, line 613: I would change the binary to hex just to keep the same types of numbers

wero1414 (changes requested):
Just add some minor changes, to allow the example to compile in any board

sabas1080:
Thanks @wero1414 , I have made the changes

wero1414:
@morganrallen @sandeepmistry this seems to fix the problems with getting random numbers with the radio, could you check it out? 

It would close the #150 PR and also the issue #394

examples/LoRandom/LoRandom.ino
```diff
@@ -63,8 +63,12 @@ void setup() {
     while (1);
   }
   Serial.print("Setting up LoRa ");
+  
+  ifdef SAMR34
```
wero1414, line 67: bad directive

wero1414 (changes requested):

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
+/*
+  LoRandomSeed
+
+  Example demonstrating how to generate random numbers using LoRa.
+
+  created 7 Sep 2020
+  by Konduino
```
tobozo, line 7: the nickname is "Kongduino" as in Hong Kong

https://github.com/Kongduino

sabas1080:
ready @wero1414 @tobozo

Kongduino:
I will do some tests, but we need to remember that the LoRa chip must be set in a specific state (continuous receive, SF 7, BW 125 KHz, C/R 4/5) before reading register 0x2c. We need to warn the user, and possibly provide a way to switch ti the proper settings, and then switch back to the user's preferred settings.

Kongduino:
See https://github.com/Kongduino/ESP32_Random_Test for an example. This version of LoRandom.h seamlessly sets up LoRa for rng, and resets LoRa to the previous settings when done.

Kongduino:
> Thanks @wero1414 , I have made the changes

There was a reason (not that important but still) to put the settings in binary: these registers have several settings within a byte, so the binary representation helps seeing the what is set up.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
+/*
+  LoRandomSeed
+
+  Example demonstrating how to generate random numbers using LoRa.
+
+  created 7 Sep 2020
+  by Kongduino
+  adapted by Andres Sabas
+  
+  SX1276 Register (address)     Register bit field (bit #)      Values    Note
```
sandeepmistry, line 10: I would suggest to remove the register notes from the example, since they are not used in the example.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+    uint8_t index = i / 16;
+    Serial.print(index, HEX); Serial.write('.');
+    Serial.println(s + t + "|");
+    s = "|"; t = "| |";
+  }
+  Serial.println(F("  +------------------------------------------------+ +----------------+"));
+}
+
+void setup() {
+  Serial.begin(115200);
```
sandeepmistry, line 53: 9600 is the default baud rate for the Arduino IDE's Serial monitor, I would suggest it be used here.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+  Serial.println("RFM_RST: " + String(RFM_RST));
+  Serial.println("RFM_DIO0: " + String(RFM_DIO0));
+  Serial.println("RFM_SWITCH: " + String(RFM_SWITCH));
+  if (!LoRa.begin(868E6)) {
+    Serial.println("Starting LoRa failed!");
+    while (1);
+  }
+  Serial.print("Setting up LoRa ");
+  
+  #ifdef SAMR34
```
sandeepmistry, line 67: Typically Arduino's examples don't use macros in sketches, please avoid them where possible.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+
+void setup() {
+  Serial.begin(115200);
+  while(!Serial);
+  Serial.print(F("\n\n\n[SX1276] Initializing ... "));
+  LoRa.setPins(SS, RFM_RST, RFM_DIO0);
+  Serial.println("SS: " + String(SS));
+  Serial.println("RFM_RST: " + String(RFM_RST));
+  Serial.println("RFM_DIO0: " + String(RFM_DIO0));
+  Serial.println("RFM_SWITCH: " + String(RFM_SWITCH));
```
sandeepmistry, line 60: Do the pin values need to be printed here?

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
+/*
```
sandeepmistry, line 1: Please add this example to the [.travis.yml](https://github.com/sandeepmistry/arduino-LoRa/blob/master/.travis.yml#L39) so it can be compiled as part of the CI.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+  Serial.println(F("  +------------------------------------------------+ +----------------+"));
+  for (uint16_t i = 0; i < len; i += 16) {
+    for (uint8_t j = 0; j < 16; j++) {
+      if (i + j >= len) {
+        s = s + "   "; t = t + " ";
+      } else {
+        char c = buf[i + j];
+        if (c < 16) s = s + "0";
+        s = s + String(c, HEX) + " ";
+        if (c < 32 || c > 127) t = t + ".";
```
sandeepmistry, line 40: Please add a comment on why the `.` is needed, it's not clear.

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+        else t = t + (char)c;
+      }
+    }
+    uint8_t index = i / 16;
```
sandeepmistry, line 44: Please avoid types not listed in https://www.arduino.cc/reference/en/. There are more used in the sketch below.

src/LoRa.cpp
```diff
@@ -607,9 +607,21 @@ void LoRaClass::setOCP(uint8_t mA)
   writeRegister(REG_OCP, 0x20 | (0x1F & ocpTrim));
 }
 
+void LoRaClass::LoRandomSeed() {
```
sandeepmistry, line 610: What do you think about calling this API `LoRa.beginRandom()` (so it's like `LoRa.beginPacket()`) - then there could also be a `LoRa.endRandom()` that could put the module in idle()?

src/LoRa.cpp
```diff
@@ -607,9 +607,21 @@ void LoRaClass::setOCP(uint8_t mA)
...
+  writeRegister(REG_OP_MODE, 0x72);
+  writeRegister(REG_MODEM_CONFIG_1, 0x72);
```
sandeepmistry, line 612: Could this re-use existing API's that set `REG_MODEM_CONFIG_1` and `REG_MODEM_CONFIG_2`? See the notes in the example sketch on what radio settings are used :)

src/LoRa.cpp
```diff
@@ -607,9 +607,21 @@ void LoRaClass::setOCP(uint8_t mA)
...
+}
+
 byte LoRaClass::random()
 {
-  return readRegister(REG_RSSI_WIDEBAND);
+  uint8_t x = 0;
+  for (uint8_t j = 0; j < 8; j++) {
+    x += (readRegister(REG_RSSI_WIDEBAND) & 0x01);
+    x = x << 1;
+    delay(1);
```
sandeepmistry, line 622: Is a delayed needed here? Does it need to be exactly 1 ms?

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+
+  To generate an N bit random number, perform N read operation of the register RegRssiWideband (address 0x2c)
+  and use the LSB of the fetched value. The value from RegRssiWideband is derived from a wideband (4MHz) signal strength
+  at the receiver input and the LSB of this value constantly and randomly changes.
+*/
+
+#include <SPI.h>
+#include <LoRa.h>
+
+void hexDump(unsigned char *buf, uint16_t len) {
```
sandeepmistry, line 28: What do you think about moving this after `loop()` to make the starting of the sketch a bit more readable?

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+  #ifdef SAMR34
+  pinMode(RFM_SWITCH, OUTPUT);
+  digitalWrite(RFM_SWITCH, 1);
+  #endif
+  
+  LoRa.setTxPower(20, PA_OUTPUT_PA_BOOST_PIN);
+  LoRa.setPreambleLength(8);
+  LoRa.LoRandomSeed();
+  Serial.println("[o]");
+  delay(1000);
```
sandeepmistry, line 76: Is this delay needed here?

examples/LoRandom/LoRandom.ino
```diff
@@ -0,0 +1,92 @@
...
+  Serial.println("End of setup\n\n");
+}
+
+void loop() {
+  unsigned char randomStock[256];
```
sandeepmistry, line 81: What do you think about renaming this variable to `randomBytes` and have it use a [byte](https://www.arduino.cc/reference/en/language/variables/data-types/byte/) type?

src/LoRa.h
```diff
@@ -81,7 +81,8 @@ class LoRaClass : public Stream {
   // deprecated
   void crc() { enableCrc(); }
   void noCrc() { disableCrc(); }
-
+  
+  void LoRandomSeed();
```
sandeepmistry, line 85: The new API needs to be added to `keywords.txt` and `API.md`. Note, I've suggested an alternative name above.

sandeepmistry (changes requested):
Hi @sabas1080, thanks for opening this pull request!

I've made some comments to start some discussions. @wero1414 @Kongduino @tobozo feel free to participate in the discussion as well.

examples/LoRandom/LoRandom.ino
Kongduino: Not necessarily. I added it to put a little more entropy.

examples/LoRandom/LoRandom.ino
Kongduino: No it was a debug for me when playing with a new board, non ESP32.

examples/LoRandom/LoRandom.ino
Kongduino: There's actually a bug in this version of hexDump. It should be:

```c
        if (c < 32 || c > 127) t = t + ".";
        else t = t + (String(c));
```

Expected output:

Screenshot 2020-09-09 at 4 19 45 PM

src/LoRa.cpp
Kongduino: That's an excellent idea.

src/LoRa.cpp
Kongduino: Sure. I tried to make LoRandom.h compatible with any LoRa lib, so we couldn't expect to have existing presets. But within LoRa.h sure.

src/LoRa.cpp
Kongduino: Not necessarily. I added it to put a little more entropy.

examples/LoRandom/LoRandom.ino
Kongduino: Not necessarily. I fail to remember why I had put it there – but I'm sure there was a reason at the time... :-)

tobozo:
Speculation : 

Use a double buffer random pool with implicit update, when the first buffer is consumed it is marked as such and updated on next opportunity. Both buffers are filled when LoRa.begin() is called, and can be separately re-filled by the driver when marked as consumed.

Benefits:

- Lora.random() actually acts as initially expected with reduced communication overhead (bonus backwards compat).
- New features to manipulate randomness (buffer size, pool size).

Kongduino:
LoRa.setPins(SS, RFM_RST, RFM_DIO0);

This is specific to a board, BastWAN, that I used originally to create LoRandom. This should definitely be removed from the example code...

src/LoRa.cpp
sabas1080: @sandeepmistry I do not understand this well, can you give me an example?

sandeepmistry: If you look at the original sketch by @Kongduino: https://github.com/sandeepmistry/arduino-LoRa/blob/4d8f70036850128ffb52ec0b3de8ab7dfe7f1b39/examples/LoRandom/LoRandom.ino#L10-L18

```
/*
  LoRandomSeed
  Example demonstrating how to generate random numbers using LoRa.
  created 7 Sep 2020
  by Konduino
  adapted by Andres Sabas
  
  SX1276 Register (address)     Register bit field (bit #)      Values    Note
  RegOpMode (0x01)              LongRangeMode[7]                ‘1’       LoRa mode enable
                                Mode[2:0]                       ‘101’     Receive Continuous mode
  ------------------------------------------------------------------------------------------------------------------
  RegModemConfig1 (0x1D)        Bw[7:4]                         ‘0111’    ‘0111’ for 125kHz modulation Bandwidth
                                CodingRate[3:1]                 ‘001’     4/5 error coding rate
                                ImplicitHeaderModeOn[0]         ‘0’       Packets have up-front header
  ------------------------------------------------------------------------------------------------------------------
  RegModemConfig2 (0x1E)        SpreadingFactor[7:4]            ‘0111’    ‘0111’ (SF7) = 6kbit/s
  To generate an N bit random number, perform N read operation of the register RegRssiWideband (address 0x2c)
  and use the LSB of the fetched value. The value from RegRssiWideband is derived from a wideband (4MHz) signal strength
  at the receiver input and the LSB of this value constantly and randomly changes.
*/
```

You can map most of the register values to:

```
setSignalBandwidth(125E3);
setCodingRate4(5);
setSpreadingFactor(7);
```

Implicit header and continuous receive mode need to be set via new methods however.

examples/LoRandom/LoRandom.ino
sandeepmistry: This still needs to be resolved ...

src/LoRa.cpp
sabas1080: I have added a new continuosMode and re-use existing API's

sabas1080:
I'm coming back to finish this task

src/LoRa.cpp
```diff
@@ -412,6 +412,11 @@ void LoRaClass::sleep()
   writeRegister(REG_OP_MODE, MODE_LONG_RANGE_MODE | MODE_SLEEP);
 }
 
+void LoRaClass::continuosMode()
```
tobozo, line 415: typo in method name ? continuosMode -> continuousMode

sabas1080: Ready

src/LoRa.h
```diff
@@ -63,6 +63,7 @@ class LoRaClass : public Stream {
 #endif
   void idle();
   void sleep();
+  void continuousMode();
```
sandeepmistry, line 66: @sabas1080 thanks for making the changes!

Things are looking very close to be ready to merge, two more notes:

1) Do you think this API should be private?

2) What do you thing about renaming this API to something like `LoRa.continuosReceiveMode()` or `LoRa.continuousRxMode()` for additional clarity?

sabas1080: @sandeepmistry ok for me, I agree

IoTThinks:
Hi,
How is the new method continuousRxMode related to the RX_CONTINOUS mode in receive() metthod?
Will the receive() use continuousRxMode?
Thanks a lot.

image

sandeepmistry:
@sabas1080 any thoughts on @IoTThinks comment in https://github.com/sandeepmistry/arduino-LoRa/pull/395#issuecomment-707183118?

plybrd:
@IoTThinks comment in https://github.com/sandeepmistry/arduino-LoRa/pull/395#issuecomment-707183118
> Hi,
> How is the new method continuousRxMode related to the RX_CONTINOUS mode in receive() metthod?
> Will the receive() use continuousRxMode?
> Thanks a lot.
> 


void LoRaClass::continuousMode()
{
  writeRegister(REG_OP_MODE, 0x72); // -> 0b10001101
}

writeRegister(REG_OP_MODE, MODE_LONG_RANGE_MODE | MODE_RX_CONTINUOUS); // ->  0b10000101

There is an description how to generate random numbers in Application Note AN1200.24 from Semtech.
See Chapter 4 of  [Random Number Generation for Cryptography](https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/440000001NAw/7YN8ZamV70_xR.vPDAAAshm.0Wt4jmRX0nOKkOzQqiI):
Bit 7 -> 1
Bit 2-0 -> 101
Nothing is said about Bit 3

Datasheet says: REGOpMode Bit 3 LowFrequencyModeOn, after Reset it is set to 1. But the setting is only important for FSK/OOK Mode.

plybrd:
Hi,
there are many application which needs only few random bytes. A big buffer of random bytes is nice only for applications which needs much of them. 
Switching between sending/receiving and generating random numbers should work on the fly.

Why not start by adding just a function
`byte LoRaClass::random(uint8_t *buffer,  size_t size)`
which gives the user just how many bytes he needs and sets the register back to the state before calling this function?

For later on this function can be used to fill some buffer or double buffer of random bytes, ...(see https://github.com/sandeepmistry/arduino-LoRa/pull/395#issuecomment-689470996).

PR #406, opened 2020-11-04
Send ACK if required or requested by the sender

brainelectronics:
Add function to send an ACK back to the sender of this message if requested.

This is of course only done if:
 * the received message is not a broadcast message
 * the sender requested an ACK from the recipient
 * this node is the addressed recipient

Some other helpful functions are added as well.
* reading all available bytes into a given buffer
* extract header from received package
* extract message from received package
* get recipient of message

### DIFF
--- a/API.md
+++ b/API.md
@@ -242,6 +242,70 @@ Returns the next byte in the packet or `-1` if no bytes are available.
 
 **Note:** Other Arduino [`Stream` API's](https://www.arduino.cc/en/Reference/Stream) can also be used to read data from the packet
 
+### Reading to buffer
+
+Read the full packet into a given buffer.
+
+```arduino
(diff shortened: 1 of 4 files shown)


IoTThinks:
Nice feature.
Just a comment on the example.

Do we really need this?
```
  // all magic happens here
  onReceive(LoRa.parsePacket());
```

Or we can just use LoRa.onReceive(onReceiveWithAck) and add in onReceiveWithAck()?
image

sandeepmistry:
Hi @brainelectronics,

Thanks for taking the time to submit this pull request!

However at this time, I would prefer not to have built-in proprietary packet header and ack support into this library. You could however create a new library that depends on this one for this functionality.

PR #425, opened 2021-01-08
fixed old URLs in readme.md

Boba0514:


IoTThinks:
Wow, I admit I never click on the URLs.
Thanks for the update.

Boba0514:
Glad I could be of benefit even if just this. :)

sandeepmistry (approved):
@Boba0514 awesome, thanks for the update!

Issue #459, opened 2021-04-20
ESP32: float value in ISR (Guru Meditation Error: Core  1 panic'ed (Coprocessor exception))


sacculus:
Hi
I'm write sketch for esp32 board with event driven model when send/receive.
I got exception when have tried print SNR & frequency error in onReceive callback like this:

`Serial.printf("% 10d: LoRa: RSSI: %d, SNR: %0.2f, Frequency error: %d\n", millis(), LoRa.packetRssi(), LoRa.packetSnr(), LoRa.packetFrequencyError());`

Exception:
`Guru Meditation Error: Core  1 panic'ed (Coprocessor exception)
Core 1 register dump:
PC      : 0x400d0fee  PS      : 0x00060a31  A0      : 0x800d17c2  A1      : 0x3ffbea50  
A2      : 0x0000153f  A3      : 0x00000001  A4      : 0x3ffbffc4  A5      : 0x00000000  
A6      : 0xffffffb7  A7      : 0x0000002f  A8      : 0x800d0fee  A9      : 0x3ffbeaf0  
A10     : 0x41080000  A11     : 0x0000005b  A12     : 0xc87536dc  A13     : 0x3ffc0794  
A14     : 0x3ffbeaac  A15     : 0x00000000  SAR     : 0x0000001d  EXCCAUSE: 0x00000004  
EXCVADDR: 0x00000000  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xfffffff4  
Core 1 was running in ISR context:
EPC1    : 0x400d0fee  EPC2    : 0x00000000  EPC3    : 0x00000000  EPC4    : 0x400841af

ELF file SHA256: 0000000000000000

Backtrace: 0x400d0fee:0x3ffbea50 0x400d17bf:0x3ffbeb10 0x400d17bf:0x3ffbeb50 0x400811e6:0x3ffbeb70 0x40081231:0x3ffbeb90 0x40083f95:0x3ffbebb0 0x400d2804:0x3ffb1fb0 0x4008689d:0x3ffb1fd0

Rebooting...
ets Jun  8 2016 00:22:57

rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[...]

PR #460, opened 2021-04-21
Issue with float in ISR on ESP32

sacculus:
Patch for [issue #459](https://github.com/sandeepmistry/arduino-LoRa/issues/459).
Methods which use float values are not allowed in ISR on ESP32 platfom because they use float coprocessor. It's possible to use double type for this.
[Link of proof](https://www.esp32.com/viewtopic.php?t=1292)


IoTThinks:
Let me check if the patch is ok for Arduino board.

PR #466, opened 2021-05-10
Set Low data rate optimization flag

d3v1c3nv11:


ve2yag:
I check lora_aprs code and this project use your library. LOWBITRATEOPTIMIZE (register config_3(0x26), bit 3) need to be set when symbol time is above 16ms.  With BW125 SF12 for APRS, the symbol time is 32.77ms and leaving this bit at 0 may cause worse decoding and bad receiver performance. Bit must be set at both side of link.

ve2yag:
Sorry, I must click on file change and you already made the change. Thank you !!

PR #570, opened 2022-03-12
Add `packetSnrRaw` method to avoid floating-point

ZacJW:
This patch is motivated by a well known (#459, #460) limitation, that ESP microcontrollers can't use `float LoRa.packetSnr()` in ISR because that method performs a floating point operation.

My proposed solution is to add a method `int LoRa.packetSnrRaw()` that returns the `REG_PKT_SNR_VALUE` register as an `int` unmodified, rather than as a float divided by 4. I've written documentation for it noting that since it forgoes the division by 4, its value is 4 times the actual SNR.

### DIFF
--- a/API.md
+++ b/API.md
@@ -204,6 +204,14 @@ float snr = LoRa.packetSnr();
 
 Returns the estimated SNR of the received packet in dB.
 
+### Packet SNR Raw
+
+```arduino
+int snrRaw = LoRa.packetSnrRaw();
+```
+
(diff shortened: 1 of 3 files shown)


torntrousers:
Could you use bit shift instead of the division - https://stackoverflow.com/questions/13577174/can-anyone-explain-why-2-shift-means-divided-by-4-in-c-codes?

ZacJW:
My thought is to leave it up to the user. I would like to keep the full precision of it, but I'd accept the loss of 2 bits.

Issue #514, opened 2021-09-29
CAD - Channel free detection?


bobricius:
Hi, Is there any way how to detect if is not transmission on channel?

Kongduino:
Yes. You need to set up `RegIrqFlagsMask`'s (0x11) bit 0 – `CadDetectedMask` – to set up Channel Activity Detection, and use `RegIrqFlags`'s (0x12) bits 0 (`CadDetected`) and 2 (`CadDone`) to decide whether the channel is free to use or not.

LoRa.h has the two bit masks defined.

```c
#define IRQ_CAD_DONE_MASK          0b00000100
#define IRQ_CAD_DETECTED_MASK      0b00000001
```
`CadDetected` is on DIO4 and `CadDone` on DIO3 in DIOx Mapping `00`, DIO1 and DIOØ in DIOx Mapping `10`. This is configured with registers `RegDioMapping1` and `RegDioMapping2` (0x40/0x41).

But there is no special provision for sending when the channel is free, like in Semtech's sx1262 library. Maybe a feature request...? This is a bit complex to set up properly.

halukmy:
any update @Kongduino ?

Kongduino:
> any update @Kongduino ?

I wasn't aware I was supposed to follow up on this... My answer is as complete as can be, given the circumstances.

halukmy:
has got any other alternatives?

morganrallen:
There are two open pull requests, if someone tests them and let's me know they work I'll merge and do a release.

bobricius:
I am implemented function from this PR ... [https://github.com/sandeepmistry/arduino-LoRa/pull/574](https://github.com/sandeepmistry/arduino-LoRa/pull/574) and work beautiful. It is logical ..."if receiving something = somebody transmitting" CAD is is almost useless, this status bit is documented in datasheet with les than one sentence

morganrallen:
That PR isn't implementing true CAD, you have to write to the `OpMode` register to enable CAD and `DioMapping1` register if you want an interrupt, as CAD is an asynchronous operation.

If I understand the `ModemStatus` register correctly, this is when the modem is in RX mode (Single, or Continuous).

So it will only detect when YOU are receiving, full CAD will detect if two OTHER devices are talking on your channel.

This functionality does look useful, but I feel like the PR is lacking, would be nice to implement all of the `ModemStatus` at the same time BUT, if someone follows up adding this to `API.md` I'll merge as is.

[...]

PR #574, opened 2022-03-27
Add LoRaClass::rxSignalDetected().

toyo:
Added function LoRaClass::rxSignalDetected() returns true when LoRa carrier is detected.

If you want to avoid collision,  you should not transmit during this function returns true.

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -18,6 +18,7 @@
 #define REG_FIFO_RX_CURRENT_ADDR 0x10
 #define REG_IRQ_FLAGS            0x12
 #define REG_RX_NB_BYTES          0x13
+#define REG_MODEM_STAT           0x18
 #define REG_PKT_SNR_VALUE        0x19
 #define REG_PKT_RSSI_VALUE       0x1a
 #define REG_RSSI_VALUE           0x1b
@@ -288,6 +289,10 @@ long LoRaClass::packetFrequencyError()
   const float fError = ((static_cast<float>(freqError) * (1L << 24)) / fXtal) * (getSignalBandwidth() / 500000.0f); // p. 37
 
   return static_cast<long>(fError);
+}
+
+bool LoRaClass::rxSignalDetected() {
+  return (readRegister(REG_MODEM_STAT) & 0x01) == 0x01;
 }
 
 int LoRaClass::rssi()
--- a/src/LoRa.h
+++ b/src/LoRa.h
@@ -44,6 +44,7 @@ public:
   int packetRssi();
   float packetSnr();
   long packetFrequencyError();
+  bool rxSignalDetected();
 
   int rssi();
 


bobricius:
Thanks! Amazing, this function I searching 2years, is much better and simpler than CAD

halukmy:
awesome! can you show some examples please? thanks

morganrallen:
Right, in addition to adding the function to `API.md` and example should be provided. It could be as simple as calling this function in the loop and displaying "Incoming message" if it returns true. In theory, a message should arrive after this becomes true.

IoTThinks:
Wow, so great Pull Request.
Hope it get reviewed and merged soon.

morganrallen:
Also trying to make some time to check this out. But also would like some clarification on this.
@toyo are you using this to detect network activity? From my reading of the datasheet I feel like this isn't working the way some people think it is. Adding an example (and update API.md) to this PR would go a long way to getting it merged.

Quoting @bobricius from issue #514 
> this status bit is documented in datasheet with les than one sentence

While correct, this sentence reads...

> The state of the LoRa modem is accessible with the ModemStatus bits in RegModemStat. They can mostly used for debug in Rx mode and the useful indicators are...

Note that it says `for debug in Rx mode`. I do not believe this method is actually used for activity detection, as I'm guessing you need to put the radio into RX mode first, so why not just use CAD(*)?

*) CAD has been merged, but I realized last night a new release hasn't been made since it was. I will make a new release in the next day or so.

toyo:
> @toyo are you using this to detect network activity?

Yes. works with SX1278.
It returns true from when it has decoded preamble to the signal end.
So, it returns false if it cannot decode preamble correctly.

morganrallen:
@toyo do you put it into receive mode first?

toyo:
> @toyo do you put it into receive mode first?

Yes, of course.

SoothingMist:
Many thanks for this new function. Added it to my baseline. Project is LoRa flood messaging. GitHub: https://github.com/SoothingMist/Scalable-Point-to-Point-LoRa-Sensor-Network.

SoothingMist:
Working on flood-messaging via LoRa P2P and using Sandeep's baseline library (https://github.com/SoothingMist/Scalable-Point-to-Point-LoRa-Sensor-Network/tree/main/Documentation%2BSoftware).

Would the following code be a worthwhile approach to CAD without modifying the library?

```
// An approach to CAD.
// Using Sandeep's baseline LoRa library:
// https://github.com/sandeepmistry/arduino-LoRa
#include <LoRa.h>
#ifdef ARDUINO_SAMD_MKRWAN1300
#error "This example is not compatible with the Arduino MKR WAN 1300 board!"
#endif

// Flags signal detection.
// OnCadDone returns false when no signal is detected.
// Returns true if a signal is detected.
uint8_t signalDetectionFlag = 02;

void setup()
{
  Serial.begin(9600);
  while (!Serial) Wait(1000);

  Serial.println("LoRa CAD Test");

  if (!LoRa.begin(915E6))
  {
    Serial.println("Starting LoRa failed!");
    Wait(1000);
    exit(1);
  }
  else Serial.println("LoRa transceiver initialized");

  // register the channel activity detection callback
  LoRa.onCadDone(onCadDone);

  // Ready
  Serial.println("Ready");
}

void loop()
{
  // try next activity detection
  signalDetectionFlag = 01;         // CAD not completed
  while(signalDetectionFlag)
  {
    signalDetectionFlag = 02;
    LoRa.channelActivityDetection();
    while(signalDetectionFlag == 02) Wait(100);
    if(signalDetectionFlag)         // signal detected
    {
      Serial.println(" detection confirmed\n");
    }
    else if( ! signalDetectionFlag) // no signal detected
    {
      Serial.println("  could send something\n");
    }
  }
  //Serial.println("This cycle completed");
  Wait(1000);
}

void onCadDone(boolean signalDetected)
{
  signalDetectionFlag = (uint8_t)signalDetected;
  if (signalDetected)
  {
    Serial.print("*** Signal Detected: "); Serial.println(signalDetected);
  }
  else
  {
    Serial.print("*** No Signal Detected: "); Serial.println(signalDetected);
  }
}

// Wait for a specific number of milliseconds.
// delay() is blocking so we do not use that.
// This approach does not use hardware-specific timers.
void Wait(long milliseconds)
{
  long beginTime = millis();
  uint8_t doSomething = 00;
  while ((millis() - beginTime) <= milliseconds) doSomething++;
}
```

SoothingMist:
Modified the earlier code to include a formal mutex. But, as I grow the project and try to include the CAD callback in a library for producing a payload and broadcasting a packet, I reach a point where the callback is never called. The library works well without CAD. So, there are no code errors or memory corruption from there. Will have to use the method suggested in this pull request. I think more has to be done with the baseline library regarding CAD.

```
// An approach to CAD.
// Using Sandeep's baseline LoRa library:
// https://github.com/sandeepmistry/arduino-LoRa
#include <LoRa.h>
#ifdef ARDUINO_SAMD_MKRWAN1300
#error "This example is not compatible with the Arduino MKR WAN 1300 board!"
#endif

// Implements formal mutex.
// https://github.com/f1rmb/Mutex
#include "Mutex.h"

// Flags signal detection.
// LoRa library's OnCadDone returns false when no signal is detected.
// Returns true if a signal is detected.
Mutex signalDetectionFlag;

// Function for displaying the amount of remaining memory.
// https://github.com/mpflaga/Arduino-MemoryFree
#include "MemoryFree.h"

void setup()
{
  Serial.begin(9600);
  while (!Serial) Wait(1000);

  Serial.println("LoRa CAD Test");

  if (!LoRa.begin(915E6))
  {
    Serial.println("Starting LoRa failed!");
    Wait(1000);
    exit(1);
  }
  else Serial.println("LoRa transceiver initialized");

  // register the channel activity detection callback
  LoRa.onCadDone(AwaitClearChannel);
  signalDetectionFlag.unlock();
  
  // Ready
  Serial.println("Ready\n");
}

void loop()
{
  // try next activity detection
  signalDetectionFlag.lock();
  LoRa.channelActivityDetection();
  while(signalDetectionFlag.isLocked()) // wait for CAD to be completed
  {
    Serial.println("locked");
    Wait(500);
  }
  Serial.println("  could send something\n");
  Wait(1000);
}

void AwaitClearChannel(boolean signalDetected)
{
  // Seeing what happens in a signal-dense situation.
  // Did not observe declining RAM.
  //Serial.print(F("Free RAM = "));
  //Serial.println(freeMemory());
  //LoRa.channelActivityDetection();
    
  if (signalDetected)
  {
    Serial.print("*** Signal Detected: "); Serial.println(signalDetected);
    LoRa.channelActivityDetection();
  }
  else
  {
    Serial.print("*** No Signal Detected: "); Serial.println(signalDetected);
    signalDetectionFlag.unlock();
  }
}

// Wait for a specific number of milliseconds.
// delay() is blocking so we do not use that.
// This approach does not use hardware-specific timers.
void Wait(long milliseconds)
{
  long beginTime = millis();
  uint8_t doSomething = 00;
  while ((millis() - beginTime) <= milliseconds) doSomething++;
}
```

PR #576, opened 2022-05-11
add setRxSingleTimeout(...)

marius-ne:
Dear LoRa Contributors,

first, thanks for your wonderful work! I implemented your library myself and was very happy with it, except that you couldn't change the timeout for the Rx Single Mode. So I went ahead and added that.

I hope you are happy with my contribution.

Best regards
Marius

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -23,6 +23,7 @@
 #define REG_RSSI_VALUE           0x1b
 #define REG_MODEM_CONFIG_1       0x1d
 #define REG_MODEM_CONFIG_2       0x1e
+#define REG_SYMB_TIMEOUT_LSB	 0x1f
 #define REG_PREAMBLE_MSB         0x20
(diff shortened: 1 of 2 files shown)


marius-ne:
Setting a timeout of 1023 symbols with a symbol duration of 1.024 ms (2^SF / BW) would result in a maximum timeout of approximately 1 second (precisely 1.048 s). After that the Timeout Interrupt would be triggered.

Kongduino:
Just one nitpick. `REG_MODEM_CONFIG_2` contains only two bits of the timeout, ie bits 9 and 1 in position 1 and 0. You should probably do something like this to be on the safe side:

```c
writeRegister(REG_MODEM_CONFIG_2, readRegister(REG_MODEM_CONFIG_2) | ((symbols >> 8)&0x03));

marius-ne:
I think that's cared for by the input check, the maximum value is 10 bits long, so the other values in REG_MODEM_CONFIG_2 will not be affected.

Issue #127, opened 2018-03-27
Lora onReceive Callback not triggered,  but data received in loop()


acbrandao:
Strange behavior, I tried altering the LoRaDuplexCallback example, integrating it into my TTGO Lora board. Sender is sending as expected, and data is reaching my receiver node, but the Callback is not getting triggered.  Code segment looks like this..

```
 setup()
{
...
  LoRa.onReceive(onReceive);  //register the Callback
  LoRa.receive();
  Serial.println("LoRa init succeeded.");
...
}

void onReceive(int packetSize) {
...
 //IT NEVER prints this line or any other code    
  Serial.println("Received packet size:"+(String)packetSize );
... 
}
```

If however I place the receive inside the main Loop() as below it works?
``` if (packetSize)     // received a packet
    {
     int  rssi = LoRa.packetRssi();
    
    Serial.print("Received packet '");
    Serial.println("Packet  RCVD "+String(packetSize)+"bytes" );
    
    // read packet
    String packet="";                   // Clear packet
    while (LoRa.available()) 
    packet += (char)LoRa.read(); // Assemble new packet    
    Serial.println(packet);
  }
```
Do I need to do anything else beside register the OnReceive for it to work?
Hardware: TTGO Lora OLED v.1

acbrandao:
Ok I tried other things like..

``` 
 LoRa.setPins(LORA_NSS, LORA_NRESET, LORA_DEFAULT_DIO0_PIN);

```
This did not help the issue, no change.

``` //pinMode(LORA_DEFAULT_DIO0_PIN, 99); ```
Caused a stack overflow  and reboot..

**Comment by torntrousers** [comment]
According to [this](https://primalcortex.wordpress.com/2017/11/24/the-esp32-oled-lora-ttgo-lora32-board-and-connecting-it-to-ttn/) for the TTGO board LORA NSS is pin 18, reset is 14 and i think DIO0 is what that refers to as LoRa-IRQ so is pin 26. So try ```LoRa.setPins(18, 14, 26)```?

**Comment by MarcRWetzel** [comment]
Hi,

I fixed this by setting the IRQ pin to INPUT. First I tested it in my own sketch, but I think this should be patched in the library, like so:

```
int LoRaClass::begin(long frequency)
{
  // setup pins
  pinMode(_ss, OUTPUT);
  pinMode(_reset, OUTPUT);
  pinMode(_dio0, INPUT); // added

[...]

```
BR
Marc

[...]

PR #621, opened 2022-12-19
Updated interrupt based blocking callback to loop based non-blocking one

sheikhmishar:
As I noticed from the source code, the onTxDone and onReceive callback is executed in an ISR. According to Arduino and ESP documentation, ISR should run nothing more than simple variable manipulation. Also, delay will not work. I have updated the code to sort out these limitations. Please accept this pull request if you think this could be helpful in time critical application

### DIFF
--- a/API.md
+++ b/API.md
@@ -127,7 +127,9 @@ Returns `1` on success, `0` on failure.
 
 ### Register callback
 
-Register a callback function for when a packet transmission finish.
+Register a callback function for when a packet transmission finish. `NOTE:` Only write simple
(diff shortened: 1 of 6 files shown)


geoffreyliddell:
This seems like a great addition! I've been using the anonymous function idea used here too and the issue is mentioned here too:  https://github.com/sandeepmistry/arduino-LoRa/issues/127  would it help this to get merged if instead of being implemented in the examples, the anonymous function were added as a default argument to the Lora.onReceive() function?

Issue #463, opened 2021-04-30
Single packet is shown as received twice


gjt211:
Board: MKRZero
LoRa Module: NiceRF SX1276-915Mhz

I have a custom board and have the following code that works except that is shows the same packet twice.
I have reduced the code to the bare minimum needed for a test.
```
#include <SPI.h>
#include <LoRa.h>

const int P_LORA_NSS = 7;
const int P_LORA_RESET = 6;
const int P_LORA_DIO0 = 10;

struct __attribute__((packed)) SENSOR_DATA11b {
  uint8_t smodel;             // 1
  uint16_t id;                // 2
  int16_t var1 = 0;           // 2
  int16_t var2 = 0;           // 2
  int16_t var3 = 0;           // 2
  int16_t crc;                // 2
};                            // 11 bytes total


SENSOR_DATA11b sData11b;

//=============================================================
void setup() {
  
  pinMode(32,OUTPUT);
  digitalWrite(32,HIGH);
  pinMode(5,OUTPUT);
  digitalWrite(5,LOW);         // Ensure mPCIE power supply is off
  pinMode(0,OUTPUT);
  digitalWrite(0,LOW);
  
  Serial.begin(115200);
  delay(2000);
  
  lora_begin();

  Serial.print(F("READY_\r"));
  Serial.println();

}

//=============================================================
void loop() {
  int packetSize = LoRa.parsePacket();
  if (packetSize == 11){
    LoRa.readBytes((uint8_t *)&sData11b, packetSize);
    Print11b(LoRa.packetRssi());
  }
}

// =============================================================
void lora_begin(){
  Serial.println(F("LoRa start"));
  LoRa.setPins(P_LORA_NSS, P_LORA_RESET, P_LORA_DIO0);
  if (!LoRa.begin(919450000)) {
    Serial.println(F("LoRa modem failure - System halted"));
    for(;;) { 
      // do nothing...
      delay(100);
    } 
  }
  LoRa.setSyncWord(0x12);
  LoRa.setSignalBandwidth(250E3); // defaults to 125E3
  LoRa.setCodingRate4(5);   // denominator of the coding rate, defaults to 5
  LoRa.setSpreadingFactor(10);
  Serial.println(F("LoRa ready"));
}


// =============================================================
void Print11b(int rssi){
  Serial.print(F(" => "));
  Serial.print(sData11b.smodel);
  Serial.print(F(","));
  Serial.print(sData11b.id);
[...]

PR #664, opened 2023-09-06
Fix duplicate packet reception on Arduino MKR 1310

mdias:
Without this change there is a chance that we receive duplicate packets

### DIFF
--- a/src/LoRa.cpp
+++ b/src/LoRa.cpp
@@ -201,6 +201,9 @@ int LoRaClass::endPacket(bool async)
     }
     // clear IRQ's
     writeRegister(REG_IRQ_FLAGS, IRQ_TX_DONE_MASK);
+#ifdef ARDUINO_SAMD_MKRWAN1310
+    writeRegister(REG_IRQ_FLAGS, IRQ_TX_DONE_MASK);
+#endif
   }
 
   return 1;
@@ -235,6 +238,9 @@ int LoRaClass::parsePacket(int size)
 
   // clear IRQ's
   writeRegister(REG_IRQ_FLAGS, irqFlags);
+#ifdef ARDUINO_SAMD_MKRWAN1310
+  writeRegister(REG_IRQ_FLAGS, irqFlags);
+#endif
 
   if ((irqFlags & IRQ_RX_DONE_MASK) && (irqFlags & IRQ_PAYLOAD_CRC_ERROR_MASK) == 0) {
     // received a packet


morganrallen:
Have you thoroughly checked the rest of your code to ensure something else isn't causing the duplicate? I've never encountered a situation where the IRQ doesn't clear.
You can test this by reading back the IRQ reg and printing out that value but note this also might fix your problem as it might just be the delay of two writes that's fixing and issue somewhere else.

mdias:
The program I have that showed this happening rather easily was basically broadcasting a counter variable every second and checking for received packets all the time.
I have 2 Arduino MKR 1310 running the same program and both of them showed like 10-20% of the packets received being received twice.

I don't have enough knowledge to investigate further why the IRQs are not clearing, but after searching online I came across [this issue](https://github.com/sandeepmistry/arduino-LoRa/issues/463) describing the same problem but on a different board, and this fix was suggested. After trying the same fix on my board I could confirm that there are no longer duplicated received packets.

I understand that this looks like a hack and the problem may come from somewhere else, so I understand if you don't want to merge it as is. But the patch as is should also only affect the MKR 1310 board which is already misbehaving (the duplicates) with the original code.

morganrallen:
There's an easy experiment you can try. Replace the new `writeRegister` call with `delay(100)`. Does this also solve the problem?

morganrallen:
It would also be helpful to understand this issue better if you could share the code around the call to `endPacket`

mdias:
Replacing `writeRegister` with `delay(100)` results in some packets not being received at all, and some others are still duplicated.

My sample application is doing this when sending a packet:
```
void SendString(const char* str)
{
  digitalWrite(LED_BUILTIN, HIGH);

  LoRa.beginPacket();
  LoRa.print(str);
  LoRa.endPacket();

  digitalWrite(LED_BUILTIN, LOW);
}
...
static void sendData()
{
  char buf[16] = {0};
  if (secondsTimer()) // trigger only once a second
  {
    sprintf(buf, "%i", counter);
    DbgPrint("sending: "); // print debug info with Serial.print()
    DbgPrintLn(buf);

    ++counter;

    SendString(buf);
  }
}
```

On the receptor end, I receive data like so:
```
bool ReceiveString(String& outString)
{
  int packetSize = LoRa.parsePacket();
  if (packetSize <= 0)
  {
    return false;
  }

  while (LoRa.available())
  {  
    outString += (char)LoRa.read();
  }

  return true;
}
```

maaaaaaaaaaaaaaaaaaaaaaaaaaax:
I have the same error on an MKR WAN 1300. I changed my code to read the IRQ flags right before and right after the call of parsePacket() and noticed that whenever a packet gets duplicated, the IRQ flags do not clear to 0 but stay at 64 after the parsePacket call.

maaaaaaaaaaaaaaaaaaaaaaaaaaax:
I got the same error on an MKR WAN 1300. I am reading the IRQ flags before and after the call of parsePacket(). Whenever a packet gets duplicated, the IRQ flags were not cleared after the call to parsePacket() but stayed at 64 instead. 

I was able to fix this by checking the content of the IRQ flag register after the return of the call to parsePacket() and clearing the register again if it is not 0 at that point. I only have to clear it once and it works immediately. If I don't clear it again, it never clears by itself, so another writing to it is necessary.

SoothingMist:
Had the same duplicate-receive problem. Employed the code changes recommended in this branch. Solved the problem. Application is LoRa flood messaging. GitHub: https://github.com/SoothingMist/Scalable-Point-to-Point-LoRa-Sensor-Network.

PR #720, opened 2025-04-26
Small updates to member functions, adding one new one.

waltje:
Moved two get/set functions fro private to public.
Added new getTxPower function.


waltje:
Messy because of some textual updates inbetween. Commit a4922b4 is the actual change.

